PR #3, opened 2016-06-06
Update post-form.php

Kapranov98:
Superfluous characters


slaFFik:
@Kapranov98 

As written in repo description: 
```
Please do not send pull requests, submit patches to 
https://buddypress.trac.wordpress.org/ instead.
```

So, please create a ticket with a patch on https://buddypress.trac.wordpress.org/ 
Thanks!

PR #6, opened 2020-05-27
Add required headers in plugin

spacedmonkey:
WordPress core supports these headers now, so a core project, we should add these headers.

### DIFF
--- a/src/bp-loader.php
+++ b/src/bp-loader.php
@@ -19,6 +19,8 @@
  * Text Domain: buddypress
  * Domain Path: /bp-languages/
  * License:     GPLv2 or later (license.txt)
+ * Requires at least: 4.8
+ * Requires PHP: 5.6
  */
 
 /**


slaFFik:
@spacedmonkey AFAIK these headers are respected when they are in `readme.txt` file only (so for wp.org plugins), and WordPress ignores them in main plugin file.
Right now these headers are [already there](https://github.com/buddypress/BuddyPress/blob/master/src/readme.txt).

spacedmonkey:
> @spacedmonkey AFAIK these headers are respected when they are in `readme.txt` file only (so for wp.org plugins), and WordPress ignores them in main plugin file.
> Right now these headers are [already there](https://github.com/buddypress/BuddyPress/blob/master/src/readme.txt).

I don't think this is correct. But also, what does it hurt to merge this?

slaFFik:
Well, https://core.trac.wordpress.org/changeset/44939 and https://core.trac.wordpress.org/changeset/44937 says WP uses API response, which uses `readme.txt` as per this PR that is merged: https://meta.trac.wordpress.org/ticket/2952 (see patch.2).

slaFFik:
And regarding "What does it hurt to merge":
As stated in repo description:
> Please do not send pull requests, submit patches to buddypress.trac.wordpress.org instead.

spacedmonkey:
For reference, it seems this was added in WordPress 5.4 here:
https://core.trac.wordpress.org/ticket/46938

PR #11, opened 2022-03-03
Add context to the 'bp_is_user_spammer' filter.

dd32:
Trac Ticket: https://buddypress.trac.wordpress.org/ticket/8645


dcavins:
Great idea. These filters are basically useless without the user object. Thanks!

PR #13, opened 2022-03-11
Removes PHP 8.1 deprecation notices and PHPUnit 9.0 errors

imath:
Trac ticket: https://buddypress.trac.wordpress.org/ticket/8649

---


composer.json
```diff
@@ -31,12 +31,12 @@
...
-		"phpcompatibility/phpcompatibility-wp": "^2.1.0",
+		"phpcompatibility/phpcompatibility-wp": "^2.1.3",
 		"dealerdirect/phpcodesniffer-composer-installer": "^0.7.1",
-		"squizlabs/php_codesniffer" : "^3.5.4",
-		"wp-coding-standards/wpcs": "*",
+		"squizlabs/php_codesniffer" : "^3.6.0",
+		"wp-coding-standards/wpcs": "~2.3.0",
 		"wp-phpunit/wp-phpunit": "^5.9",
-		"phpunit/phpunit": "^7.5",
+		"phpunit/phpunit": "^9.0",
```
renatonascalves, line 39: I'd recommend removing this line altogether. Let's leave `yoast/phpunit-polyfills` to decide which version to install based on the current PHP version.

`composer.lock` should also be removed from source control so that we don't lock packages per PHP version.

renatonascalves, line 39: This is blocking WordPress 5.7 to be tested properly in the Github Action CI (we don't see the failed actions because they are not active for pull requests). We need to install the `"phpunit/phpunit": "^7.5"` version in the Github Action only for WordPress 5.7 or previous WordPress versions (right now, we are only testing 5.7)

src/bp-groups/bp-groups-functions.php
```diff
@@ -2969,7 +2969,7 @@ function bp_groups_get_group_types( $args = array(), $output = 'names', $operato
 		$types = bp_get_taxonomy_types( bp_get_group_type_tax_name(), $types );
 	}
 
-	$types = wp_filter_object_list( $types, $args, $operator );
+	$types = array_filter( wp_filter_object_list( $types, $args, $operator ) );
```
renatonascalves, line 2972: This concern is related to all those array_filter in this pull request.

Should we run `array_values` as well?

> Array keys are preserved, and may result in gaps if the array was indexed. The result array can be reindexed using the [array_values()](https://www.php.net/manual/en/function.array-values.php) function.

Source: https://www.php.net/manual/en/function.array-filter.php

tests/phpunit/includes/testcase-emails.php
```diff
@@ -3,8 +3,8 @@
 
 class BP_UnitTestCase_Emails extends BP_UnitTestCase {
 
-	public static function setUpBeforeClass() {
-		parent::setUpBeforeClass();
+	public static function set_up_before_class() {
```
renatonascalves, line 6: 🎈

renatonascalves (changes requested):
I'm happy to collaborate on some of those suggestions. :)

Nice work!

composer.json
```diff
@@ -31,12 +31,12 @@
...
 		"yoast/phpunit-polyfills": "^1.0.1"
 	}
```
renatonascalves, line 41: Could we also add support to composer v2?

```
"config": {
	"allow-plugins": {
		"composer/installers": true,
		"dealerdirect/phpcodesniffer-composer-installer": true
	}
}
```

renatonascalves:
Almost forgot the `bootstrap.php` file will also need to have this introduced:

```
// Setting PHPUnit polyfills.
const WP_TESTS_PHPUNIT_POLYFILLS_PATH = __DIR__ . '/../../vendor/yoast/phpunit-polyfills';
```

src/bp-groups/bp-groups-functions.php
imath: I don't think so. I'm using it to remove empty values from the array.

composer.json
imath: Sure, I'll add it 👍

composer.json
imath: Ok, I've removed the line completely as you suggested in the previous comment.

imath: Ok, I've removed the line 👍

src/bp-groups/bp-groups-functions.php
renatonascalves: That's correct but if those values were an array with indexes, the result array index will not be the same. We'd be essentially change the structure of the array.

imath:
@renatonascalves Thanks a lot for your recommandations. I've applied them and ran tests locally using PHP 7.4.30 and PHP 8.1.9. In both cases everything went fine. Could you have another look at it?

composer.json
imath: I've also removed composer.lock

renatonascalves:
@imath Could you also add support for the Github Action to run in pull requests?

src/bp-groups/bp-groups-functions.php
imath: Right, but I prefer to preserve keys than losing them in favor of a numerically indexed array

> `array_values()` returns all the values from the array and indexes the array **numerically**.

imath:
@renatonascalves Just did, I've also included PHP 8.1 into the main PHP matrix.

renatonascalves:
@imath Cool! I think the last task is adding support for PHPUnit here: https://github.com/buddypress/buddypress/blob/master/.github/workflows/unit-tests.yml#L61-L62

`"phpunit/phpunit": "^7.5",`

src/bp-groups/bp-groups-functions.php
renatonascalves: 🤔 

Maybe what I'm saying is that you are **losing the order** of the index using `array_filter`, which might be unexpected?! Or is that expected?

imath: I'm unsure these arrays are numerically indexed. I'm beat for tonight! I need some rest. WP 5.7 is still problematic: unit tests are running with PHPUnit 7.5 but are failing 😢 . Let's look at it later!

renatonascalves: If they are not numerically indexed, this suggestion can be skipped. Just wanted to raise that to make sure. :)

renatonascalves:
Looks like there is an actual bug in the unit tests. 🤔

rafiahmedd:
image
Hi,

I just run the test on the master branch, however, I couldn't find any issues on the test. Am I doing anything wrong here?

renatonascalves:
@rafiahmedd Are you using WordPress 5.7?

imath:
@renatonascalves yes there are bugs related to WordPress 5.7. I believe it's due to the fact `wp-phpunit/wp-phpunit:5.7` doesn't include WordPress changes about using Yoast polifyier ((I believe these were included in 5.9).

I think @rafiahmedd tested on BuddyPress master, not this PR's branch 😉.

I'm wondering if it wouldn't fail if we were simply removing these lines into our GH action: https://github.com/buddypress/buddypress/blob/412b003469beef19af079e428f0e40ee5bd09887/.github/workflows/unit-tests.yml#L61|L63

It should use `wp-phpunit/wp-phpunit:6.0` which includes Yoast polifyier... I'll try to give it a try locally.

renatonascalves:
I'd think it was going to fail since WordPress 5.7 won't won't work with `"phpunit/phpunit": "^7.5"`. But maybe it will work with `wp-phpunit/wp-phpunit:6.0`.

It's worth a test.

rafiahmedd:
@imath yeah you are right I tested this with the master branch. I will check this from this branch.

imath:
Touchdown 🏈  @renatonascalves I was right about my intuition, using the `wp-phpunit/wp-phpunit` and PHPunit version for all tests fixes the failing tests under WP 5.7. I believe the risk of doing so is very limited as we only rely on WP Unit Tests class / factory and not on testcases. What's your opinion about it?

rafiahmedd:
The latest changes working perfectly. I think we can go with this one.

renatonascalves (approved):
Nice! All is good here!

imath:
Awesome many thanks to you both @renatonascalves & @rafiahmedd I'll commit changes on trunk asap 🍾🥂

PR #19, opened 2022-06-15
Fix wrong LIMIT offset value

marklawntalk:
The current LIMIT offset calculator is wrong resulting in duplicate acitivty cards.
Possible related to https://buddypress.trac.wordpress.org/ticket/4535

Fix the correct formula to calculate the 'offset' value.

Trac ticket: https://buddypress.trac.wordpress.org/ticket/4535
---

### DIFF
--- a/src/bp-notifications/classes/class-bp-notifications-notification.php
+++ b/src/bp-notifications/classes/class-bp-notifications-notification.php
@@ -482,7 +482,7 @@ class BP_Notifications_Notification {
 		if ( ! empty( $args['page'] ) && ! empty( $args['per_page'] ) ) {
 			$page     = absint( $args['page']     );
 			$per_page = absint( $args['per_page'] );
-			$offset   = $per_page * ( $page - 1 );
+			$offset   = $per_page * ( $page - 1 ) + 1;
 			$retval   = $wpdb->prepare( "LIMIT %d, %d", $offset, $per_page );
 		}
 


imath:
Hi @marklawntalk are you sure it's about activity cards as your patch is editing the notifications' table offset limit query ?

marklawntalk:
Hi @imath , my bad, yes it's the notification module. The site I worked on uses the notifications module for both activity and notification modules hence the confusion. But I understand their difference.

imath:
So here's what in place.

```php
$page     = absint( $args['page'] );
$per_page = absint( $args['per_page'] );
$offset   = $per_page * ( $page - 1 );
$retval   = $wpdb->prepare( "LIMIT %d, %d", $offset, $per_page );
```

Let's take some exemples.

1. To get the first ten results, we need the 1st page of results which should be converted to `LIMIT 0, 10` for MySql (as the offset of the initial row is 0 (not 1), see https://dev.mysql.com/doc/refman/8.0/en/select.html) : 
- `$args['per_page']` is 10,
-  `$args['page']` is 1.

```php
$page     = 1;
$per_page = 10;
$offset   = 10 * ( 1 - 1 ); // = 0
$retval   = $wpdb->prepare( "LIMIT %d, %d", 0, 10 ); // LIMIT 0, 10
```
✅ Which is right imho.

2. If we need to get the second page of results, it should be converted to : `LIMIT 10, 20` for MySql

- `$args['per_page']` is 10,
-  `$args['page']` is 2.

```php
$page     = 2;
$per_page = 10;
$offset   = 10 * ( 2 - 1 ); // = 10
$retval   = $wpdb->prepare( "LIMIT %d, %d", 10, 20 ); // LIMIT 10, 20
```
✅ Which is right imho.

With your suggested patch, we'd have a wrong offset value, because we would ignore the first db entry each time : 
```php
$page     = 1;
$per_page = 10;
$offset   = 10 * ( 1 - 1 ) + 1; // = 1
$retval   = $wpdb->prepare( "LIMIT %d, %d", 1, 10 ); // LIMIT 1, 10
```
❌ Which is wrong imho.

So everything is fine with how the offset is generated. Thanks for your report, but I'm going to close it and leave the code the way it is.

PR #21, opened 2022-08-14
Set up PHPCS

renatonascalves:
Trac ticket: https://buddypress.trac.wordpress.org/ticket/7228

---

### DIFF
--- a/.github/workflows/unit-tests.yml
+++ b/.github/workflows/unit-tests.yml
@@ -55,7 +55,7 @@ jobs:
         run: composer validate --strict
 
       - name: Install Composer dependencies
-        uses: ramsey/composer-install@v1
+        uses: ramsey/composer-install@v2
         with:
           composer-options: "--ignore-platform-reqs"
 
@@ -87,13 +87,13 @@ jobs:
           npm run wp-env start -- --update
 
       - name: Running single site unit tests
-        run: npm run test-php
         if: ${{ success() || failure() }}
+        run: npm run test-php
 
       - name: Running multi site unit tests
-        run: npm run test-php-multisite
         if: ${{ success() || failure() }}
+        run: npm run test-php-multisite
 
       - name: Stop Docker environment
-        run: npm run wp-env stop
         if: ${{ success() || failure() }}
+        run: npm run wp-env stop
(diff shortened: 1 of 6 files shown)


.github/workflows/unit-tests.yml
```diff
@@ -87,13 +87,13 @@ jobs:
           npm run wp-env start -- --update
 
       - name: Running single site unit tests
-        run: npm run test-php
```
imath, line 90: Correct me if I'm wrong, this means PHPUnit tests won't run if PHPCS fails, right?

I thought we were not being too strict about PHPCS at first. I believe we should leave PHPUnit tests run even if PHPCS Fails.

If I'm wrong, why moving the `run` command under the `if` one ?

imath (commented):
Thanks for your work on this. I need some explanations about the unit-tests.yml changes. I left a comment inside the inline code.

.github/workflows/unit-tests.yml
renatonascalves: Nope, they are not related.

I changed the order just so that it aligns with the expected order in the Github Action documentation, that being:

- name
- if
- run

The order of the operation is not important for _how_ the actions are run and this doesn't mean unit tests will only run if PHPCS fails.

Those if conditional are there in case actions in this workflow fail. Not actions for another workflow (aka, phpcs).

renatonascalves: I actually think we should remove the `failure()` status but I don't know why it was introduced in the first place. Maybe there are good reasons. 🤔

renatonascalves: Since we are using a `fail-fast` strategy, the `failure()` status would never be called.

imath: Awesome, thanks a lot for these explanations. Let's leave `failure()` the way it is. It's now time to commit these improvements on https://buddypress.trac.wordpress.org/ (don't forget we don't merge PR on this repo). While committing on SVN, you can use the trick @dd32 shared with us here: https://bpdevel.wordpress.com/2022/03/06/using-pr-to-share-your-patches-on-buddypress-trac/#comment-31467

If you got unused to SVN (eg: making sure to use `svn add new-file` before `svn commit` when a new file has been added...) and would feel more comfortable if I'll handle the commit, don't hesitate to ping me 😉

renatonascalves: Cool! I'll commit to SVN later tonight. :)

PR #31, opened 2022-10-27
Performance improvements to the `BP_Messages_Thread` class

renatonascalves:
As titled.

In this first part, I'm making sure data is present in a performant way. The second part is changing `get_messages` and `get_recipients` to stop getting all data at once.

Currently, if a thread has 10k messages, `get_messages` will try to get all messages in a single SQL query. The same is true for `get_recipients`. This is a breaking change, so it bears more discussion.

Trac ticket: https://buddypress.trac.wordpress.org/ticket/8697
Trac ticket: https://buddypress.trac.wordpress.org/ticket/8696

---

### DIFF
--- a/src/bp-messages/classes/class-bp-messages-thread.php
+++ b/src/bp-messages/classes/class-bp-messages-thread.php
@@ -1,6 +1,6 @@
 <?php
 /**
- * BuddyPress Messages Classes.
+ * BuddyPress Messages Thread Class.
  *
  * @package BuddyPress
  * @subpackage MessagesClasses
@@ -26,12 +26,20 @@ class BP_Messages_Thread {
 	public $thread_id;
 
 	/**
-	 * The current messages.
+	 * The current messages in the message thread.
 	 *
 	 * @since 1.0.0
 	 * @var array
 	 */
 	public $messages;
+
+	/**
+	 * The current messages count in the message thread.
+	 *
+	 * @since 11.0.0
+	 * @var int
+	 */
+	public $messages_count;
 
 	/**
 	 * The current recipients in the message thread.
@@ -40,6 +48,14 @@ class BP_Messages_Thread {
 	 * @var array
 	 */
 	public $recipients;
+
+	/**
+	 * The current recipients count in the message thread.
+	 *
+	 * @since 11.0.0
+	 * @var int
+	 */
+	public $recipients_count;
 
 	/**
 	 * The user IDs of all messages in the message thread.
@@ -189,19 +205,27 @@ class BP_Messages_Thread {
 			return false;
 		}
 
-		$last_message_index         = count( $this->messages ) - 1;
-		$this->last_message_id      = $this->messages[ $last_message_index ]->id;
-		$this->last_message_date    = $this->messages[ $last_message_index ]->date_sent;
-		$this->last_sender_id       = $this->messages[ $last_message_index ]->sender_id;
-		$this->last_message_subject = $this->messages[ $last_message_index ]->subject;
-		$this->last_message_content = $this->messages[ $last_message_index ]->message;
+		// Messages count.
+		$this->messages_count = self::get_total_thread_message_count( $this->thread_id );
 
-		foreach ( (array) $this->messages as $key => $message ) {
-			$this->sender_ids[ $message->sender_id ] = $message->sender_id;
+		// Sender IDs.
+		$this->sender_ids = self::get_thread_sender_ids( $this->thread_id );
+
+		// Get and set last message.
+		$message = self::get_latest_thread_message( $this->thread_id );
+		if ( ! empty( $message ) && is_object( $message ) ) {
+			$this->last_message_id      = $message->id;
+			$this->last_message_date    = $message->date_sent;
+			$this->last_sender_id       = $message->sender_id;
+			$this->last_message_subject = $message->subject;
+			$this->last_message_content = $message->message;
 		}
 
 		// Fetch the recipients and set the displayed/logged in user's unread count.
 		$this->recipients = $this->get_recipients( $thread_id, $r );
+
+		// Set total recipients count.
+		$this->recipients_count = self::get_total_thread_recipients_count( $this->thread_id );
 
 		// Grab all message meta.
 		if ( true === (bool) $r['update_meta_cache'] ) {
@@ -214,8 +238,8 @@ class BP_Messages_Thread {
 		 * @since 2.2.0
 		 * @since 10.0.0 Added `$r` as a parameter.
 		 *
-		 * @param BP_Messages_Thread $this Message thread object.
-		 * @param array              $r    Array of paremeters.
+		 * @param BP_Messages_Thread $message_thread Message thread object.
+		 * @param array              $r              Array of paremeters.
 		 */
 		do_action( 'bp_messages_thread_post_populate', $this, $r );
 	}
@@ -252,7 +276,7 @@ class BP_Messages_Thread {
 	 * @global BuddyPress $bp The one true BuddyPress instance.
 	 * @global wpdb $wpdb WordPress database object.
 	 *
-	 * @param int   $thread_id Message thread ID.
+	 * @param int   $thread_id The message thread ID.
 	 * @param array $args      {
 	 *     Array of arguments.
 	 *     @type int|null $recipients_page     Page of recipients being requested. Default to all.
@@ -386,6 +410,7 @@ class BP_Messages_Thread {
 				)
 			);
 
+			// Cast integers.
 			foreach ( $messages as $key => $data ) {
 				$messages[ $key ]->id        = (int) $messages[ $key ]->id;
 				$messages[ $key ]->thread_id = (int) $messages[ $key ]->thread_id;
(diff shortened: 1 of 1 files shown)


src/bp-messages/classes/class-bp-messages-thread.php
```diff
@@ -419,12 +444,63 @@ public static function get_messages( $thread_id = 0, $args = array() ) {
...
+	 * @global BuddyPress $bp The one true BuddyPress instance.
+	 * @global wpdb $wpdb WordPress database object.
+	 *
+	 * @param int $thread_id The message thread ID.
+	 * @return object|null
+	 */
+	public static function get_latest_thread_message( $thread_id ) {
+		global $wpdb;
+
+		$message = wp_cache_get( $thread_id, 'bp_messages_thread_latest_message' );
```
imath, line 461: This new cache group should probably be setup into this function:
https://github.com/buddypress/buddypress/blob/master/src/bp-messages/classes/class-bp-messages-component.php#L417|L428

I believe this was the initial purpose of: https://buddypress.trac.wordpress.org/ticket/8590

src/bp-messages/classes/class-bp-messages-thread.php
```diff
@@ -942,7 +1021,121 @@ public static function get_total_threads_for_user( $user_id, $box = 'inbox', $ty
...
+	 * @global BuddyPress $bp The one true BuddyPress instance.
+	 * @global wpdb $wpdb WordPress database object.
+	 *
+	 * @param int $thread_id The message thread ID.
+	 * @return int Total thread recipients count.
+	 */
+	public static function get_total_thread_recipients_count( $thread_id ) {
+		global $wpdb;
+
+		$total_recipients_count = wp_cache_get( $thread_id, 'bp_messages_recipients_count' );
```
imath, line 1080: This new cache group should probably be setup into this function:
https://github.com/buddypress/buddypress/blob/master/src/bp-messages/classes/class-bp-messages-component.php#L417|L428

I believe this was the initial purpose of: https://buddypress.trac.wordpress.org/ticket/8590

src/bp-messages/classes/class-bp-messages-thread.php
```diff
@@ -942,7 +1021,121 @@ public static function get_total_threads_for_user( $user_id, $box = 'inbox', $ty
...
+	 * @global wpdb $wpdb WordPress database object.
+	 *
+	 * @param int $thread_id The message thread ID.
+	 * @return array Total sender ids.
+	 */
+	public static function get_thread_sender_ids( $thread_id ) {
+		global $wpdb;
+
+		$bp         = buddypress();
+		$sender_ids = wp_cache_get( $thread_id, 'bp_messages_sender_ids' );
```
imath, line 1118: This new cache group should probably be setup into this function:
https://github.com/buddypress/buddypress/blob/master/src/bp-messages/classes/class-bp-messages-component.php#L417|L428

I believe this was the initial purpose of: https://buddypress.trac.wordpress.org/ticket/8590

src/bp-messages/classes/class-bp-messages-thread.php
```diff
@@ -189,20 +205,28 @@ public function populate( $thread_id = 0, $order = 'ASC', $args = array() ) {
...
+		// Messages count.
+		$this->messages_count = self::get_total_thread_message_count( $this->thread_id );
 
-		foreach ( (array) $this->messages as $key => $message ) {
-			$this->sender_ids[ $message->sender_id ] = $message->sender_id;
+		// Sender IDs.
+		$this->sender_ids = self::get_thread_sender_ids( $this->thread_id );
+
+		// Get and set last message.
+		$message = self::get_latest_thread_message( $this->thread_id );
```
imath, line 215: If I understand well this part: you suggest to add a new query to get the latest thread's message instead of finding the latest thread's message that should be into the `$this->messages`? 
Is this a downsize of introducing messages' thread pagination?
Could we avoid this new query, making sure to include the latest thread's message ID, for example ?

src/bp-messages/classes/class-bp-messages-thread.php
```diff
@@ -189,20 +205,28 @@ public function populate( $thread_id = 0, $order = 'ASC', $args = array() ) {
...
-		$this->last_sender_id       = $this->messages[ $last_message_index ]->sender_id;
-		$this->last_message_subject = $this->messages[ $last_message_index ]->subject;
-		$this->last_message_content = $this->messages[ $last_message_index ]->message;
+		// Messages count.
+		$this->messages_count = self::get_total_thread_message_count( $this->thread_id );
 
-		foreach ( (array) $this->messages as $key => $message ) {
-			$this->sender_ids[ $message->sender_id ] = $message->sender_id;
+		// Sender IDs.
+		$this->sender_ids = self::get_thread_sender_ids( $this->thread_id );
```
imath, line 212: This introduces a new query, while we used to use `$this->messages` to set the list of `sender_ids` right?
Is this a downsize of introducing messages' thread pagination?

src/bp-messages/classes/class-bp-messages-thread.php
```diff
@@ -189,20 +205,28 @@ public function populate( $thread_id = 0, $order = 'ASC', $args = array() ) {
...
+			$this->last_sender_id       = $message->sender_id;
+			$this->last_message_subject = $message->subject;
+			$this->last_message_content = $message->message;
 		}
 
 		// Fetch the recipients and set the displayed/logged in user's unread count.
 		$this->recipients = $this->get_recipients( $thread_id, $r );
 
+		// Set total recipients count.
+		$this->recipients_count = self::get_total_thread_recipients_count( $this->thread_id );
```
imath, line 228: Another query, I guess it's still a downsize of pagination?

imath (changes requested):
Hi @renatonascalves 

Thanks a lot for your work on this. If I understand well, your suggestion is to introduce 3 new db queries to be more consistent when paginating messages or users of a thread.

Could we only run these new queries only if we request pagination ?

renatonascalves:
@imath Yes!

Only running this when paginated is not good enough IMO. We are still querying ALL messages on the database (or in the cache) by default (**regardless if the request is paginated or not**) which is exactly what I'm trying to solve as part of this task.

This is a flaw in the design which I postponed solving when I introduced messages/recipients pagination but that ultimately should be solved regardless (at some point, why not now?!). The performance issue is more obvious when you have apps with threads with thousands of messages or/and recipients. (I've been doing some stress testing with the BP GraphQL extension.)

I'm aware this might be a breaking change and I'm open to options for how we can smooth this update to users. But doing those requests separately will be more performant (since they are cached) IF we get a smaller subset of messages. And since the data is not related to the current batch of messages, I can't see how we could get them without doing separate, cached, queries.

The second part of this task is to actually stop the default behavior from querying ALL messages /recipients by default and set a reasonable number. Allowing the option to get all messages/recipients, like `posts_per_page = -1`.

--

There is an argument to be made here that those two new `messages|recipients_count` MySQL queries should not be part of this class. We could move them to the BP REST API if desired. But if we follow through on this update, we'll need them regardless.

src/bp-messages/classes/class-bp-messages-thread.php
renatonascalves: No, this is a new piece of data that I'm introducing. The Slack app is a good example of how this information might be useful. Currently, we return the recipients `array`, but since it is paginated, the user never gets the total number of recipients. They need to request all and calculate from that.

src/bp-messages/classes/class-bp-messages-thread.php
renatonascalves: Currently, we are requesting ALL messages in a thread. With the change to remove this behavior `$this->messages` won't have all messages so we need another way to get the sender ids.

src/bp-messages/classes/class-bp-messages-thread.php
renatonascalves: Not related to the pagination. This is related to the default behavior of requesting ALL messages by default. Which is not very performant for apps with thousands of messages.

src/bp-messages/classes/class-bp-messages-thread.php
renatonascalves: Good cache!

imath:
@renatonascalves 

Thanks for your explanations, here's what I suggest as it might be a breaking change as you wrote and we're ~8 days before 11.0.0-beta1:

1. Make a first step about returning the total number of messages per thread to cover [#BPTrac-8697](https://buddypress.trac.wordpress.org/ticket/8697) (which I'm not sure you did into this PR), if the messages of the thread are paginated or if a specific argument is passed eg: what we're doing [here for the Activity API](https://github.com/buddypress/buddypress/blob/master/src/bp-activity/classes/class-bp-activity-activity.php#L379).
2. Make another step about making sure to return the latest message of the thread, your `self::get_latest_thread_message( $this->thread_id );` to cover [#BPTrac-8696](https://buddypress.trac.wordpress.org/ticket/8696)  in case the the messages of the thread are paginated, otherwise we already have it.
3. Open a new ticket on BP Trac for the other performance improvements, and share data/benchmarks about it to explain/prove how 4 queries are taking less time than 1. Ideally, we'd need to commit these kind of improvements early during a development cycle so that we have enough time to test it. Moreover we should use these improvements into Core, eg: adding pagination into the Template packs (<- I can help on this part if needed).

What do you think ?

renatonascalves:
Sounds like an awesome plan! I'll bump related tickets to the next version so that we have more time to implement and test those changes. o/

PR #48, opened 2022-12-28
PhpDoc: adding global param for `$bp`

renatonascalves:
- As titled;
- Removing `$bp` when not used on the function/method;
- Minor tweaks.

---


imath (changes requested):
Hi @renatonascalves Thanks a lot for your work on this. I believe we shouldn't add `@global` for `buddypress()` as it's not available into `$GLOBALS` so it's a bit misleading strictly speaking.

PR #51, opened 2022-12-30
Get the total number of messages of a thread for `BP_Messages_Thread`

renatonascalves:
Trac ticket: https://buddypress.trac.wordpress.org/ticket/8697

---

### DIFF
--- a/src/bp-messages/classes/class-bp-messages-thread.php
+++ b/src/bp-messages/classes/class-bp-messages-thread.php
@@ -26,12 +26,20 @@ class BP_Messages_Thread {
 	public $thread_id;
 
 	/**
-	 * The current messages.
+	 * The current messages in the message thread.
 	 *
 	 * @since 1.0.0
 	 * @var array
 	 */
 	public $messages;
+
+	/**
+	 * The current messages count in the message thread.
+	 *
+	 * @since 12.0.0
+	 * @var int
+	 */
+	public $messages_total_count;
 
 	/**
 	 * The current recipients in the message thread.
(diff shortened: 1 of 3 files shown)


renatonascalves:
We could also not take this approach at all and do the actual MySQL query in the BP REST API plugin. Here I'm assuming this new piece of data might be useful for those using or extending `BP_Messages_Thread`.

src/bp-messages/classes/class-bp-messages-thread.php
```diff
@@ -945,6 +956,43 @@ public static function get_total_threads_for_user( $user_id, $box = 'inbox', $ty
...
+	 * @global BuddyPress $bp The one true BuddyPress instance.
+	 * @global wpdb $wpdb WordPress database object.
+	 *
+	 * @param int $thread_id The message thread ID.
+	 * @return int Total thread message count
+	 */
+	public static function get_total_thread_message_count( $thread_id ) {
+		global $wpdb;
+
+		$total_count = wp_cache_get( $thread_id, 'bp_messages_unread_count' );
```
imath, line 973: I think we should use another cache group, as this one is used to get the unread count. This function, if I understand well, gets the amount of messages for a thread no matter if messages are unread or read. About the cache group name, something like `bp_messages_thread_count` might be suitable...

NB: You'll need to also add the new cache group to the `BP_Messages_Component::setup_cache_groups()`

imath (changes requested):
Thanks for working on this. I have a concern about how the count is cached. We should use a specific cache group instead of one being used for something else (get unread messages count). If you can add unit tests about cache, that would be terrific 😄

src/bp-messages/classes/class-bp-messages-thread.php
renatonascalves: Ops! Good cache! This is a mistake!

renatonascalves:
@imath Thanks for the cache reminder. I updated the code to use a different approach, including support for cache busting. :)

imath (commented):
Awesome 👏. Ready to be committed on SVN trunk 👌

PR #55, opened 2023-01-10
PHPDoc: changes to action callbacks that should not be returning anything

renatonascalves:
There are a lot of action callbacks being misused.

This pull request:

- removes the `@return` PHPdoc in functions without returned values, only with side effects;
- remove boolean values from functions without returned values, only with side effects;
- Update `add_action` to `add_filter` where applicable.

---

### DIFF
--- a/src/bp-core/classes/class-bp-admin-types.php
+++ b/src/bp-core/classes/class-bp-admin-types.php
@@ -72,8 +72,6 @@ class BP_Admin_Types {
 	 * Register BP Types Admin.
 	 *
 	 * @since 7.0.0
-	 *
-	 * @return BP_Admin_Types
 	 */
 	public static function register_types_admin() {
 		if ( ! is_admin() ) {
@@ -86,7 +84,7 @@ class BP_Admin_Types {
 			$bp->core->types_admin = new self;
 		}
 
-		return $bp->core->types_admin;
+		$bp->core->types_admin;
 	}
 
 	/**
@@ -459,7 +457,6 @@ class BP_Admin_Types {
 	 *
 	 * @param WP_Term $term     The term object for the BP Type.
 	 * @param string  $taxonomy The type taxonomy name.
-	 * @return string           HTML Output.
 	 */
 	public function edit_form_fields( $term = null, $taxonomy = '' ) {
 		if ( ! isset( $term->name ) || ! $term->name || ! $taxonomy ) {
@@ -482,7 +479,7 @@ class BP_Admin_Types {
 			}
 		}
 
-		return $this->add_form_fields( $taxonomy, $type );
+		$this->add_form_fields( $taxonomy, $type );
 	}
 
 	/**
--- a/src/bp-messages/classes/class-bp-messages-notices-admin.php
+++ b/src/bp-messages/classes/class-bp-messages-notices-admin.php
@@ -43,8 +43,6 @@ class BP_Messages_Notices_Admin {
 	 * Create a new instance or access the current instance of this class.
 	 *
 	 * @since 3.0.0
-	 *
-	 * @return BP_Messages_Notices_Admin
 	 */
 	public static function register_notices_admin() {
 
@@ -58,7 +56,7 @@ class BP_Messages_Notices_Admin {
 			$bp->messages->admin = new self;
 		}
 
-		return $bp->messages->admin;
+		$bp->messages->admin;
 	}
 
 	/**
--- a/src/bp-notifications/bp-notifications-functions.php
+++ b/src/bp-notifications/bp-notifications-functions.php
@@ -471,13 +471,10 @@ function bp_notifications_delete_notifications_from_user( $user_id, $component_n
  * @since 2.5.0
  *
  * @param int $user_id ID of the user who is about to be deleted.
+ * @return int|bool
  */
 function bp_notifications_delete_notifications_on_user_delete( $user_id ) {
-	if ( 'wpmu_delete_user' !== current_action() ) {
-		_doing_it_wrong( __FUNCTION__, __( 'This function is only used once a user is deleted. Please use `bp_notifications_delete_notifications_from_user()` instead.', 'buddypress' ), '12.0.0' );
-	}
-	
-	BP_Notifications_Notification::delete( array(
+	return BP_Notifications_Notification::delete( array(
 		'user_id'           => $user_id,
 		'item_id'           => false,
 		'secondary_item_id' => false,
@@ -485,7 +482,18 @@ function bp_notifications_delete_notifications_on_user_delete( $user_id ) {
 		'component_name'    => false,
 	) );
 }
-add_action( 'wpmu_delete_user', 'bp_notifications_delete_notifications_on_user_delete' );
+
+/**
+ * Callback: Delete a user's notifications when the user is deleted.
+ *
+ * @since 12.0.0
+ *
+ * @param int $user_id ID of the user who is about to be deleted.
+ */
+function bp_notifications_delete_notifications_on_user_delete_callback( $user_id ) {
+	bp_notifications_delete_notifications_on_user_delete( $user_id );
+}
+add_action( 'wpmu_delete_user', 'bp_notifications_delete_notifications_on_user_delete_callback' );
 
 /**
  * Deletes user notifications data on the 'delete_user' hook.
--- a/src/bp-notifications/bp-notifications-functions.php
+++ b/src/bp-notifications/bp-notifications-functions.php
@@ -471,7 +471,7 @@ function bp_notifications_delete_notifications_from_user( $user_id, $component_n
  * @since 2.5.0
  *
  * @param int $user_id ID of the user who is about to be deleted.
- * @return int|bool
+ * @return int|false The number of rows deleted, or false on error.
  */
 function bp_notifications_delete_notifications_on_user_delete( $user_id ) {
 	return BP_Notifications_Notification::delete( array(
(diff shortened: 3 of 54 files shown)


src/bp-core/classes/class-bp-admin-types.php
```diff
@@ -86,7 +84,7 @@ public static function register_types_admin() {
 			$bp->core->types_admin = new self;
 		}
 
-		return $bp->core->types_admin;
+		$bp->core->types_admin;
```
imath, line 87: This change is making me nervous, I'd like to be sure it doesn't break the Types Admin first by testing it.

src/bp-messages/classes/class-bp-messages-notices-admin.php
```diff
@@ -58,7 +56,7 @@ public static function register_notices_admin() {
 			$bp->messages->admin = new self;
 		}
 
-		return $bp->messages->admin;
+		$bp->messages->admin;
```
imath, line 59: Same thing here, I'm a bit nervous about this change. I need to be sure it doesn't break the notices management screen.

src/bp-notifications/bp-notifications-functions.php
```diff
@@ -471,10 +471,9 @@ function bp_notifications_delete_notifications_from_user( $user_id, $component_n
  * @since 2.5.0
  *
  * @param int $user_id ID of the user who is about to be deleted.
- * @return int|false The number of rows deleted, or false on error.
  */
 function bp_notifications_delete_notifications_on_user_delete( $user_id ) {
-	return BP_Notifications_Notification::delete( array(
+	BP_Notifications_Notification::delete( array(
```
imath, line 476: This change is making me nervous, my concern is if a plugin is using this function to actually get a returned value it might generate issues with it. Could we include a check to be sure the current action is the hook it's attached to, in other words that it's not being use directly ?

src/bp-xprofile/classes/class-bp-xprofile-user-admin.php
```diff
@@ -38,7 +36,7 @@ public static function register_xprofile_user_admin() {
 			$bp->profile->admin = new self;
 		}
 
-		return $bp->profile->admin;
+		$bp->profile->admin;
```
imath, line 39: Once again, I'd like to test this change to make sure it doesn't break the xProfile WP Admin screen.

imath (commented):
Thanks for your work on this. I have concerns about some changes especially the ones that are returning an instance of a class. I believe these needs to be manually tested.

renatonascalves:
@imath I did some quick preliminary testing locally and didn't see any issues.

If you could check/test as well, that'd be great. 💯 

> my concern is if a plugin is using this function to actually get a returned value it might generate issues with it.

I agree! I have other ways (more verbose) to address the PHPDoc issues without changing the examples you raised.

But I still think those examples are incorrect, they should not be returning anything, since they are being used to register/instantiate classes to BuddyPress itself. So if anyone is using them, they are doing it incorrectly.


`bp_notifications_delete_notifications_on_user_delete` is a good example. It currently has a side effect, AND it has a returned value. But when hooking into the `do_action`, the returned value has no usage. 

This I think was a flaw in the design (its usage). Making a getter/setter function and using it in a `do_action` hook.

src/bp-notifications/bp-notifications-functions.php
renatonascalves: Let's create a different function with the side effect instead of bundling those two together since they serve different purposes.

src/bp-messages/classes/class-bp-messages-notices-admin.php
renatonascalves: I tested this and didn't see any issues or warnings.

src/bp-core/classes/class-bp-admin-types.php
renatonascalves: tested this and didn't see any issues or warnings.

src/bp-xprofile/classes/class-bp-xprofile-user-admin.php
imath: Just tested this. I saw no side effects. It's could to go.

src/bp-messages/classes/class-bp-messages-notices-admin.php
imath: Just tested this. I saw no side effects. It's could to go.

src/bp-core/classes/class-bp-admin-types.php
imath: Just tested this. I saw no side effects. It's could to go.

src/bp-notifications/bp-notifications-functions.php
imath: Ok, this is what I suggest, it the current action is not corresponding to the hook this function is attached to > throw a doing it wrong.

```suggestion
	if ( 'wpmu_delete_user' !== current_action() ) {
		_doing_it_wrong( __FUNCTION__, __( 'This function is only used once a user is deleted. Please use `bp_notifications_delete_notifications_from_user()` instead.', 'buddypress' ), '12.0.0' );
	}
	
	BP_Notifications_Notification::delete( array(
```

Could you also update the `bp_notifications_delete_notifications_from_user()` function so that its arguments have default values, see below: 

```php
function bp_notifications_delete_notifications_from_user( $user_id = 0, $component_name = false, $component_action = false ) {
```

imath:
Hi @renatonascalves 

I tested things I had a concerned about. No side effects, so it's good to go. About `bp_notifications_delete_notifications_on_user_delete()` I've added a suggestion to deal with it. Thanks again for your work on it.

src/bp-notifications/bp-notifications-functions.php
renatonascalves: @imath If I wanted to use this function in a custom/new hook, this new `doing_it_wrong` error would stop me.

imath: Right! Thanks for your feedback. We're actually directly using this function in `bp_notifications_delete_notifications_on_delete_user()`. Here's what I would suggest for now. Let's leave this part the way it is for now, commit the rest of the PR, and open a specific ticket about it to find the best way to deal with it.

renatonascalves: I mean, we are doing, and will do more of, this change in several places of the codebase. Why are you singling this example out for discussion?

We should do just like the other examples, remove the `return` and let those affected adapt to this change. Unless you are planning to discuss every single instance since there has been a lot, and I'm not even done with several components yet, so I'm sure they will be more.

imath: Here's how I see things: if I see a potential problem with plugins about a change, I will try to prevent it. That's one less than all the ones I haven't seen. If being in line with WPCS means potentially break a plugin, then I choose not to be in line with WPCS and I simply add `// phpcs: ignore`.

What you say is: there's a potential problem with plugins, but nevermind they will adapt. First I'm pretty sure they wouldn't, and in the end the end users will complain in forums and we'll loose time to understand what's going wrong.

So we can try to find the best way to deal with it, or you can commit, and I will revert this particular change, open a ticket to try to find the best way to deal with it.

BTW, it looks like you committed my suggestion into your PR according to failing tests. Weird.

renatonascalves: I see some misunderstandings in your feedback, so I'm suggesting moving the conversation about PHPDoc and WPCS to [Slack](https://wordpress.slack.com/archives/C02RQBYUG/p1674399913706989) so that I can clarify what I meant there.

For the actual change, we can keep the function with the same name and create a different one for the `do_action` hook. The name of the function with the returned value is misleading though, `bp_notifications_delete_notifications_on_user_delete`, since it was created to be run in a `do_action`. But this approach will account for your concerns, in case a plugin author is using it.

renatonascalves:
@imath Could you give another pass/look?

src/bp-notifications/bp-notifications-functions.php
```diff
@@ -482,7 +482,18 @@ function bp_notifications_delete_notifications_on_user_delete( $user_id ) {
...
-add_action( 'wpmu_delete_user', 'bp_notifications_delete_notifications_on_user_delete' );
+
+/**
+ * Callback: Delete a user's notifications when the user is deleted.
+ *
+ * @since 12.0.0
+ *
+ * @param int $user_id ID of the user who is about to be deleted.
+ */
+function bp_notifications_delete_notifications_on_user_delete_callback( $user_id ) {
```
renatonascalves, line 493: We can also name it `bp_notifications_delete_notifications_on_user_delete_action`. I see other functions in other components with similar nomemclature.

imath: Awesome! Thanks a lot for your work on this. I like your solution 💪. I'm fine with both function names.

imath (commented):
Thanks @renatonascalves, great job. Feel free to commit on SVN trunk 👍

src/bp-notifications/bp-notifications-functions.php
imath: Sure, your solution is fine with me 💯

imath (changes requested):
Hi @renatonascalves we've discussed with @dcavins about this change and we think we shouldn't change returned value for these functions but instead create specific functions to use as action callbacks and put them into a `bp-component-hooks.php` file: see https://bpdevel.wordpress.com/2023/02/04/bp-dev-chat-summary-january-30-2023/#hooks-callback

PR #56, opened 2023-01-14
PHPDoc: fixing several issues in the Blogs component

renatonascalves:

---

### DIFF
--- a/src/bp-blogs/bp-blogs-activity.php
+++ b/src/bp-blogs/bp-blogs-activity.php
@@ -350,7 +350,7 @@ function bp_blogs_format_activity_action_new_blog_comment( $action, $activity ) 
 		$comment = get_comment( $activity->secondary_item_id );
 		restore_current_blog();
 
-		if ( ! empty( $comment ) && is_a( $comment, 'WP_Comment' ) ) {
+		if ( ! empty( $comment ) && $comment instanceof WP_Comment ) {
 			$action = apply_filters_deprecated( 'bp_blogs_activity_new_comment_action', array( $action, $comment, $post_url . '#' . $activity->secondary_item_id ), '2.0.0', 'bp_blogs_format_activity_action_new_blog_comment' );
 		}
 	}
--- a/src/bp-blogs/bp-blogs-activity.php
+++ b/src/bp-blogs/bp-blogs-activity.php
@@ -238,7 +238,7 @@ function bp_blogs_format_activity_action_new_blog_post( $action, $activity ) {
 		$post = get_post( $activity->secondary_item_id );
 		restore_current_blog();
 
-		if ( ! empty( $post ) && is_a( $post, 'WP_Post' ) ) {
+		if ( ! empty( $post ) && $post instanceof WP_Post ) {
 			$post_url = add_query_arg( 'p', $post->ID, trailingslashit( get_home_url( $activity->item_id ) ) );
 			$action   = apply_filters_deprecated( 'bp_blogs_activity_new_post_action', array( $action, $post, $post_url ), '2.0.0', 'bp_blogs_format_activity_action_new_blog_post' );
 		}
--- a/src/bp-blogs/bp-blogs-activity.php
+++ b/src/bp-blogs/bp-blogs-activity.php
@@ -14,8 +14,6 @@ defined( 'ABSPATH' ) || exit;
  * Register activity actions for the blogs component.
  *
  * @since 1.0.0
- *
- * @return bool|null Returns false if activity component is not active.
  */
 function bp_blogs_register_activity_actions() {
 	if ( is_multisite() ) {
@@ -240,7 +238,7 @@ function bp_blogs_format_activity_action_new_blog_post( $action, $activity ) {
 		$post = get_post( $activity->secondary_item_id );
 		restore_current_blog();
 
-		if ( ! empty( $post ) && ! is_wp_error( $post ) ) {
+		if ( ! empty( $post ) && is_a( $post, 'WP_Post' ) ) {
 			$post_url = add_query_arg( 'p', $post->ID, trailingslashit( get_home_url( $activity->item_id ) ) );
 			$action   = apply_filters_deprecated( 'bp_blogs_activity_new_post_action', array( $action, $post, $post_url ), '2.0.0', 'bp_blogs_format_activity_action_new_blog_post' );
 		}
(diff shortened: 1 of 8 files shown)


src/bp-blogs/bp-blogs-activity.php
```diff
@@ -240,7 +238,7 @@ function bp_blogs_format_activity_action_new_blog_post( $action, $activity ) {
 		$post = get_post( $activity->secondary_item_id );
 		restore_current_blog();
 
-		if ( ! empty( $post ) && ! is_wp_error( $post ) ) {
+		if ( ! empty( $post ) && is_a( $post, 'WP_Post' ) ) {
```
imath, line 241: I believe it's best to use `instanceof`.

```suggestion
		if ( ! empty( $post ) && $post instanceof WP_Post ) {
```

src/bp-blogs/bp-blogs-activity.php
```diff
@@ -352,7 +350,7 @@ function bp_blogs_format_activity_action_new_blog_comment( $action, $activity )
 		$comment = get_comment( $activity->secondary_item_id );
 		restore_current_blog();
 
-		if ( ! empty( $comment ) && ! is_wp_error( $comment ) ) {
+		if ( ! empty( $comment ) && is_a( $comment, 'WP_Comment' ) ) {
```
imath, line 353: I believe it's best to use `instanceof`.

```suggestion
		if ( ! empty( $comment ) && $comment instanceof WP_Comment ) {
```

src/bp-blogs/bp-blogs-functions.php
```diff
@@ -1415,11 +1409,11 @@ function bp_blogs_add_blogmeta( $blog_id, $meta_key, $meta_value, $unique = fals
  * Remove all blog associations for a given user.
  *
  * @param int $user_id ID whose blog data should be removed.
- * @return bool Returns false on failure.
  */
 function bp_blogs_remove_data( $user_id ) {
-	if ( !is_multisite() )
-		return false;
+	if ( !is_multisite() ) {
```
imath, line 1414: Let's add a space between ! & the function 😉

```suggestion
	if ( ! is_multisite() ) {
```

src/bp-blogs/classes/class-bp-blogs-template.php
```diff
@@ -163,8 +163,6 @@ public function __construct( ...$args ) {
 		// Set the total blog count.
 		if ( empty( $r['max'] ) || ( $r['max'] >= (int) $this->blogs['total'] ) ) {
 			$this->total_blog_count = (int) $this->blogs['total'];
-		} else {
-			$this->total_blog_count = (int) $max;
```
imath, line 167: Why do you remove the default `total_blog_count` value ?

src/bp-blogs/classes/class-bp-blogs-template.php
```diff
@@ -176,8 +174,6 @@ public function __construct( ...$args ) {
 		// Set the current blog count.
 		if ( empty( $r['max'] ) || ( $r['max'] >= (int) $blog_count ) ) {
 			$this->blog_count = (int) $blog_count;
-		} else {
-			$this->blog_count = (int) $max;
```
imath, line 180: Why do you remove the default `blog_count` value ?

imath (changes requested):
Hi @renatonascalves Thanks a lot for your work here 👍. I've added some suggestions & I have a concern about two blog loop property where you removed default value. I'd like to understand why 😅.

src/bp-blogs/bp-blogs-functions.php
renatonascalves: I plan on tackling this in the whole project soon. 💪🏾

src/bp-blogs/classes/class-bp-blogs-template.php
renatonascalves: If you look at the code, you will see that the $max variable does not exist. This code is defaulting to 0 since `(int) null === 0`. Since my previous changes, the `$total_blog_count` property already defaults to 0, so this is not necessary.

src/bp-blogs/classes/class-bp-blogs-template.php
renatonascalves: If you look at the code, you will see that the `$max` variable does not exist. This code defaults to 0 since `(int) null === 0`. Since my previous changes, the `$blog_count` property already defaults to 0, so this is unnecessary.

src/bp-blogs/classes/class-bp-blogs-template.php
renatonascalves: Source: https://github.com/buddypress/buddypress/blob/989ed99592aea76bafb0ffac9a5b32520603974c/src/bp-blogs/classes/class-bp-blogs-template.php#L180

src/bp-blogs/classes/class-bp-blogs-template.php
renatonascalves: Source: https://github.com/buddypress/buddypress/blob/989ed99592aea76bafb0ffac9a5b32520603974c/src/bp-blogs/classes/class-bp-blogs-template.php#L180

src/bp-blogs/classes/class-bp-blogs-template.php
imath: Then it's probably a bug that happened when one of us removed the use of the `extract()` function. If you look at this line https://github.com/buddypress/buddypress/blob/989ed99592aea76bafb0ffac9a5b32520603974c/src/bp-blogs/classes/class-bp-blogs-template.php#L130 It's possible to use a `max` argument in blogs queries, so `$max` should be `(int) wp_unslash( $r['max'] )`.

src/bp-blogs/classes/class-bp-blogs-template.php
imath: As explained in my previous comment, we should make sure $max is defined instead of removing it. Eg:

`$max = (int) wp_unslash( $r['max'] );`

src/bp-blogs/classes/class-bp-blogs-template.php
renatonascalves: 👍🏾

imath (commented):
Looks good! Thanks a lot for these improvements. Feel free to commit it on SVN Trunk 👍

PR #73, opened 2023-03-11
5th step of the BP Rewrites merge: make more BuddyPress generated links use BP Rewrites

imath:
- Improve the Members component adding permastructs and custom rewrite rules for registration and activation pages.
- Improve the Blogs component adding custom rewrite rule to handle the Blogs create page.
- Make a huge progress about replacing all occurrences of `bp_get_root_domain()` (45 replacements were performed). There might be some last ones inside PHPUnit tests.
- Deprecate `bp_displayed_user_domain()`, `bp_loggedin_user_domain()`, `bp_groups_directory_permalink()`, `bp_get_groups_directory_permalink()`,  `bp_blogs_directory_permalink()` & `bp_get_blogs_directory_permalink()` and replace them with these new functions: `bp_displayed_user_url()`, `bp_loggedin_user_url()`, `bp_groups_directory_url()`, `bp_get_groups_directory_url()`, `bp_blogs_directory_url()` & `bp_get_blogs_directory_url()`. These deprecations are required because these functions were used to build some BuddyPress URLs concatenating URL chunks.
- Start replacing `bp_loggedin_user_domain()` & `bp_displayed_user_domain()` with the new functions.

Trac ticket: https://buddypress.trac.wordpress.org/ticket/4954

---

### DIFF
--- a/src/bp-activity/bp-activity-functions.php
+++ b/src/bp-activity/bp-activity-functions.php
@@ -3262,10 +3262,18 @@ function bp_activity_get_permalink( $activity_id, $activity_obj = false ) {
 	if ( false !== array_search( $activity_obj->type, $use_primary_links ) ) {
 		$link = $activity_obj->primary_link;
 	} else {
-		if ( 'activity_comment' == $activity_obj->type ) {
-			$link = bp_get_root_domain() . '/' . bp_get_activity_root_slug() . '/p/' . $activity_obj->item_id . '/#acomment-' . $activity_obj->id;
+		$path_chunks = array(
+			'component_id'                 => 'activity',
+			'single_item_action'           => 'p',
+			'single_item_action_variables' => array( $activity_obj->id ),
+		);
+
+		if ( 'activity_comment' === $activity_obj->type ) {
+			$path_chunks['single_item_action_variables'] = array( $activity_obj->item_id );
+
+			$link = bp_rewrites_get_url( $path_chunks ) . '#acomment-' . $activity_obj->id;
 		} else {
-			$link = bp_get_root_domain() . '/' . bp_get_activity_root_slug() . '/p/' . $activity_obj->id . '/';
+			$link = bp_rewrites_get_url( $path_chunks );
 		}
 	}
 
--- a/src/bp-activity/bp-activity-template.php
+++ b/src/bp-activity/bp-activity-template.php
@@ -86,6 +86,11 @@ function bp_activity_directory_permalink() {
 	 * @return string Activity directory permalink.
 	 */
 	function bp_get_activity_directory_permalink() {
+		$url = bp_rewrites_get_url(
+			array(
+				'component_id' => 'activity',
+			)
+		);
 
 		/**
 		 * Filters the activity directory permalink.
@@ -94,7 +99,7 @@ function bp_activity_directory_permalink() {
 		 *
 		 * @param string $url Permalink url for the activity directory.
 		 */
-		return apply_filters( 'bp_get_activity_directory_permalink', trailingslashit( bp_get_root_domain() . '/' . bp_get_activity_root_slug() ) );
+		return apply_filters( 'bp_get_activity_directory_permalink', $url );
 	}
 
 /**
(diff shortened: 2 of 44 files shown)


sbrajesh:
This is a real bad idea. This is going to create havoc as the functions you are trying to deprecate are used very commonly. 
Why not change the internal code of the functions to use current method for generating url instead of deprecating? what are the cons of doing that?
If the current deprecation spree keeps going on with the core functions, this will kill the whole BuddyPress eco system(not all themes and plugins are going to update their code for the trivial reason of name change).
Please reconsider it.

imath:
Hi,

Thanks for your feedback. It's nice to read from you, it's been a while. If you can reply with a solution to this question I'm very open to reconsider it.

How can you know when a plugin author uses this code: `bp_blogs_directory_permalink() . 'create/'`? I need to make sure to convert it to an URL built using the BP Rewrites so that it can use the customized slug and be reachable using pretty or plain links ?

sbrajesh:
Hi,
Thank you for your work on the BuddyPress core! 
It is nice to participate again.

As far as I am aware BuddyPress has limited number of end point which allow create actions(group,blogs).
If we look at these 2, we only have to worry about these 2 end points with hard coded action. A simple solution will be to add a canonical redirection from old url to new and have a filter to toggle this behaviour for these two.

For everything else, there are proper functions available in BuddyPress that points to the url. My main concerns were bp_loggedin_user_domain, bp_displayed_user_domain which are pervasive. These were marked deprecated in this commit
https://github.com/buddypress/buddypress/pull/73/commits/5da133d58cf97310ce697471bafd892713220cac
I am looking forward to spend time with the rewrite changes this week and have a better grasp to provide any feedback/assistance from next week.

imath:
I wish it there was "proper functions available in BuddyPress that points to the URL".

I've spent 2 years on the BP Rewrites plugin and I can tell you it's not the case. Here are some examples from our code base:
- ​​bp_displayed_user_domain() . bp_get_profile_slug() . '/change-avatar/delete-avatar
- bp_loggedin_user_domain() . bp_get_members_invitations_slug() . '/send-invites'
- ​​bp_displayed_user_domain() . bp_get_messages_slug() . '/' . bp_current_action()

There are a lot more.

So you're suggesting to redirect all hardcoded URLs?

sbrajesh:
Great points.
In my opinion, a function should be deprecated if the stop doing what they are supposed to do.
How a function is used by others may not be a ground for it. Even the newly introduced functions can be misused and hard coded by some people.

I will put one example:-
bp_loggedin_user_domain() - The purpose of this function is to provide url of the logged user. Has that changed? is that not achievable anymore by internally modifying it? If the purpose did not change, there is no rationale behind deprecating it.
Same goes for others.
Core can not track all the use cases and can not stop the misuse but the core can update the function to reflect the current approach and also replace the hard coded instance in the core itself(if there are some).
The goal should be to allow users minimum friction to update.

Even with the new rewrite API, the end points are not going to change automatically unless the admin does it manually. Making them aware that they test their change for url and report to the breaking plugin/theme will save a lot of time and resentment from community.
Also, when I said redirect, my intention was for group/blog creation only. I don't see any harm in doing it from other end points as part of a back compat plugin if enough people demand it.
 
Thank you for your attention.

imath:
@sbrajesh 

Thanks for your feedback. I'm more in line with this definition of [deprecation](https://techterms.com/definition/deprecated). These 2 functions will not be used anymore in BuddyPress once BP Rewrites will be fully merged. As we used to use them the wrong way, concatenating URL chunks, plugin authors did the same and it's fine. We need to stop doing so in favor of new functions making sure it's possible to customize slugs, be compatible with plain links and comply with a WordPress standard: the WP Rewrite API. I believe our role is to tell plugin developers they need to update their plugin for community benefits. When a user will have an issue he won't look for the code responsible for the fail, he will write a topic into our support forum giving more work to people contributing to support.

We agreed on this scenario:
- move to BP Rewrites
- build a BP Classic plugin for plugin who won't update their code.

My plan is to deprecate the problematic functions to be sure thanks to notices I replaced these functions from our code base.

Then I can move these functions into the BP Classic plugin. If this plugin is not used the deprecation notices will help us to inform people are using a plugin requiring the BP Classic plugin to be compatible with BuddyPress.

If I don't have this deprecation notices, I'm losing this possibility.

As @dcavins and you seem very concerned about the two functions (logged in & displayed domains). I'm going to leave them the way they are for now (although I think it will end up generating issues). It will make my work more difficult. Then we'll discuss again about these 2 functions. So far your arguments haven't convinced me, but I need to move on.

I'll commit this PR soon.

sbrajesh:
Hi @imath 
Thank you.
I am sure I know a bit or two about deprecation. My concern is your reason behind deprecation.
In order to deprecate something, you have to think about its impact and the benefits. is it absolutely necessary? 

When you allow 3rd parties to use your API, you need to consider it even more deeply than if you were using your own APIs.
I am sure you have good intention but the direction and rationale is plainly incorrect. It is like knowingly breaking every existing BuddyPress theme and plugin(yes, throwing notice is same as breaking).
With the current shrinking market that BuddyPress has I doubt that there will be a recovery from this move. It will be much prudent for developers to work with platforms which provide stability like WordPress does instead of working with a volatile API.

It is your decision as you have put a lot of work in BuddyPress but I will suggest that you re-asses the way WordPress and the APIs WordPress eco system works. When did you remember a major API break in WordPress? There was a time around 10 -12 years ago when a change in $wpdb->query had a wide impact but other than that, it has always been trivial change for existing applications.
This is how a software that has a large eco system/dependent plugin systems work. APIs are like contract and you don't break because you got a new one. you break it when it can't be maintained.

Anyway, I am not a big fan of breaking existing things for the sake of changing names and what you are doing is going outside the scope of the core to force others to update. Not sure if that will help or be harmful for the BP.

With that I will leave it for you to take the final decision.

PR #165, opened 2023-09-15
Initiate /docs/releases/ folder + test content formatting from the old codex to the new doc

jaz-on:
- Initiate the /docs/releases/ folder.
- Copying https://codex.buddypress.org/releases/version-9-2-0/ post from the old Codex as an example.
- Adding Markdown formatting to the copied content.


imath (commented):
Hi @jaz-on 

Thanks a lot for this PR. I'm taking this opportunity to explain a few more things about how to get ready to use the `docs` first level subdirectories as Handbooks on a WordPress Site.

1. We need a manifest.json to map `md` files to WP posts
2. File names shouldn't contain spaces and should use lower case
3. I'm not sure we want a "Releases" handbook

Where do you think releases should live in user, developer or contributor handbooks ?

PR #219, opened 2024-01-18
Make sure to check for deprecated function files according to the current major release

imath:
- On install, the initial version is always set to the closest major one, so there's no need to get it again as the `$first_major` variable.
- Instead of comparing initial major version with the current version, we need to use the current **major** version so that minor releases also includes the right deprecated function files.
- Improve the unit test to include a check with a current minor version.

Trac ticket: https://buddypress.trac.wordpress.org/ticket/9075

---

### DIFF
--- a/src/bp-core/bp-core-functions.php
+++ b/src/bp-core/bp-core-functions.php
@@ -4846,30 +4846,30 @@ function bp_get_deprecated_functions_versions() {
 
 	// List of versions containing deprecated functions.
 	$deprecated_functions_versions = array(
-		'1.2',
-		'1.5',
-		'1.6',
-		'1.7',
-		'1.9',
-		'2.0',
-		'2.1',
-		'2.2',
-		'2.3',
-		'2.4',
-		'2.5',
-		'2.6',
-		'2.7',
-		'2.8',
-		'2.9',
-		'3.0',
-		'4.0',
-		'6.0',
-		'7.0',
-		'8.0',
-		'9.0',
-		'10.0',
-		'11.0',
-		'12.0',
+		1.2,
+		1.5,
+		1.6,
+		1.7,
+		1.9,
+		2.0,
+		2.1,
+		2.2,
+		2.3,
+		2.4,
+		2.5,
+		2.6,
+		2.7,
+		2.8,
+		2.9,
+		3.0,
+		4.0,
+		6.0,
+		7.0,
+		8.0,
+		9.0,
+		10.0,
+		11.0,
+		12.0,
 	);
 
 	/*
@@ -4893,8 +4893,8 @@ function bp_get_deprecated_functions_versions() {
 	 * 2 last versions deprecated functions will be loaded for upgraded installs.
 	 */
 	$initial_version        = (float) bp_get_initial_version();
-	$current_version        = (float) bp_get_version();
-	$load_latest_deprecated = $initial_version < bp_get_major_version( $current_version );
+	$current_major_version  = (float) bp_get_major_version( bp_get_version() );
+	$load_latest_deprecated = $initial_version < $current_major_version;
 
 	// New installs.
 	if ( ! $load_latest_deprecated ) {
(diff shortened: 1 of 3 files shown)


src/bp-core/bp-core-functions.php
```diff
@@ -4894,7 +4894,7 @@ function bp_get_deprecated_functions_versions() {
 	 */
 	$initial_version        = (float) bp_get_initial_version();
 	$current_version        = (float) bp_get_version();
-	$load_latest_deprecated = $initial_version < $current_version;
+	$load_latest_deprecated = $initial_version < bp_get_major_version( $current_version );
```
emaralive, line 4897: **`bp_get_major_version( $current_version )`** returns a **string** value, e.g., "12.0", and this line is making a comparison between a float value (**`$initial_version`**) and the aforementioned **string** value.  Is this what you want to do?

vapvarun: You are correct; comparing a float value ($initial_version) and a string value ($current_version) is inappropriate and may lead to unexpected behavior or incorrect results.
$current_major_version  = (float) bp_get_major_version( $current_version );
$load_latest_deprecated = $initial_version < $current_major_version;

imath: Many thanks to you both, I'll update the PR asap

src/bp-core/bp-core-functions.php
```diff
@@ -4920,13 +4920,7 @@ function( $file ) {
...
 	}
 
-	// Try to get the first major version that was in used when BuddyPress was fist installed.
-	$first_major = '';
-	if ( $initial_version ) {
-		$first_major = bp_get_major_version( $initial_version );
-	}
-
-	$index_first_major = array_search( $first_major, $deprecated_functions_versions, true );
+	$index_first_major = array_search( $initial_version, $deprecated_functions_versions, true );
```
emaralive, line 4923: Here is another **`$initial_version`** is a **float** value and **`$deprecated_functions_versions`** is an **array of strings**.

src/bp-core/bp-core-functions.php
```diff
@@ -4937,7 +4931,7 @@ function( $file ) {
 		$latest_deprecated_functions_versions = array_splice( $latest_deprecated_functions_versions, -2 );
 	}
 
-	$index_initial_version = array_search( $first_major, $latest_deprecated_functions_versions, true );
+	$index_initial_version = array_search( $initial_version, $latest_deprecated_functions_versions, true );
```
emaralive, line 4934: And,, yet, another **`$initial_version`** is a float value and **`$latest_deprecated_functions_versions`** is an array of strings (looks like it would be a single element with a value of "12.0", if I'm not mistaken).

vapvarun: $latest_deprecated_functions_versions_float = array_map('floatval', $latest_deprecated_functions_versions);
$index_initial_version = array_search($initial_version, $latest_deprecated_functions_versions_float, true);

Instead of making these point-of-use changes, we can also initially check and make the $deprecated_functions_versions array float.

emaralive: True about apples to apples comparisons, at some point the types should be made the same.

src/bp-core/bp-core-functions.php
```diff
@@ -4937,7 +4931,7 @@ function( $file ) {
...
 	if ( false !== $index_initial_version ) {
```
emaralive, line 4935: I believe, this is going to be a problem because, **`$index_initial_version`** is most likely going to be an integer value of 0 (zero), if I'm not mistaken.

src/bp-core/bp-core-functions.php
imath: Thanks for this one too 👍

src/bp-core/bp-core-functions.php
imath: I think it's ok, 0 or another integer value means array contains the initial version so we can remove it.

src/bp-core/bp-core-functions.php
imath: I choose to use an array of float values instead.

src/bp-core/bp-core-functions.php
imath: It should be ok now the list of versions is an array of float values.

src/bp-core/bp-core-functions.php
imath: It should be ok now the list of versions is an array of float values.

PR #223, opened 2024-01-22
Attributes in block.json files are not translated

imath:
1) Fix the Latest Activities block:
- Avoid a React Error removing the code used to set the postId attribute. We now use specific attributes for the ServerSideRender component.
- Fix a regressions introduced by [r13537](https://buddypress.trac.wordpress.org/changeset/13537/) that was preventing the usage of the `activity/widget` template when BP Nouveau is the active template pack.
- Fix a misusage of `wp_list_pluck()` preventing activity types to be fetched for the block.

2) For BP Blocks using a `title` attributes, make sure the default one can be translated.
- Latest Activities block,
- Recent Networkwide Posts
- Dynamic Groups List
- Recently Active Members
- Dynamic Members List
- Online Members

Trac ticket: https://buddypress.trac.wordpress.org/ticket/9074

---


dcavins:
Wow, the way this works is very counterintuitive.

imath:
The main problem was with the Latest Activities block, something that used to work previously (setting the postId attribute once the block is added to a post) is now generating an error preventing the Activity Types select box to appear etc.. So I had to use a workaround 😓 . 

I noticed it while working on a way to have the default Blocks title translated. Using Block attributes in the block.json file was a wrong idea. Title, description are translated but not attributes.

PR #228, opened 2024-02-12
Improve BP debug info

imath:
Use JavaScript to make sure the BP help tab link is only added when the BuddyPress help tab is active.

bp-debug-info

Trac ticket: https://buddypress.trac.wordpress.org/ticket/9093

---

### DIFF
--- a/src/bp-core/admin/bp-core-admin-tools.php
+++ b/src/bp-core/admin/bp-core-admin-tools.php
@@ -757,6 +757,10 @@ add_action( 'network_admin_notices', 'bp_core_admin_notice_repopulate_blogs_resu
 function bp_core_admin_debug_information( $debug_info = array() ) {
 	global $wp_settings_fields;
 
+	if ( ! bp_is_root_blog() ) {
+		return $debug_info;
+	}
+
 	$active_components = wp_list_pluck( bp_core_get_active_components( array(), 'objects' ), 'name', 'id' );
 	$bp_settings       = array();
 	$skipped_settings  = array( '_bp_theme_package_id', '_bp_community_visibility' );
--- a/src/bp-core/admin/bp-core-admin-tools.php
+++ b/src/bp-core/admin/bp-core-admin-tools.php
@@ -756,10 +756,11 @@ add_action( 'network_admin_notices', 'bp_core_admin_notice_repopulate_blogs_resu
  */
 function bp_core_admin_debug_information( $debug_info = array() ) {
 	global $wp_settings_fields;
-	$active_components    = array_intersect_key( bp_core_get_components(), buddypress()->active_components );
-	$bp_settings          = array();
-	$non_numeric_settings = array( '_bp_theme_package_id', '_bp_community_visibility' );
-	$bp_url_parsers       = array(
+
+	$active_components = wp_list_pluck( bp_core_get_active_components( array(), 'objects' ), 'name', 'id' );
+	$bp_settings       = array();
+	$skipped_settings  = array( '_bp_theme_package_id', '_bp_community_visibility' );
+	$bp_url_parsers    = array(
 		'rewrites' => __( 'BP Rewrites API', 'buddypress' ),
 		'legacy'   => __( 'Legacy Parser', 'buddypress' ),
 	);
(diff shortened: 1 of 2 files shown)


src/bp-core/admin/bp-core-admin-tools.php
```diff
@@ -756,8 +756,14 @@ function bp_core_admin_notice_repopulate_blogs_resume() {
  */
 function bp_core_admin_debug_information( $debug_info = array() ) {
 	global $wp_settings_fields;
-	$active_components = array_intersect_key( bp_core_get_components(), buddypress()->active_components );
+
+	if ( ! bp_is_root_blog() ) {
```
renatonascalves, line 760: What's the rationale behind this decision? If I had a network of blogs (BP instances), I'd expect to have data specific to each BP instance in each site, no?

src/bp-core/admin/bp-core-admin-tools.php
```diff
@@ -813,31 +832,103 @@ function bp_core_admin_debug_information( $debug_info = array() ) {
...
 }
 add_filter( 'debug_information', 'bp_core_admin_debug_information' );
+
+/**
+ * Adds a BuddyPress section to the Site Health Info Admin Screen help tabs.
+ *
+ * @since 14.0.0
+ */
+function bp_core_admin_debug_information_add_help_tab() {
+	if ( ! bp_is_root_blog() ) {
```
renatonascalves, line 884: Same question.

renatonascalves (approved):
🌵

src/bp-core/admin/bp-core-admin-tools.php
imath: Interesting @renatonascalves to me there can only be 1 BP instance no matter the multisite config, as there's only 1 users table and as https://github.com/buddypress/buddypress/blob/master/src/bp-core/bp-core-options.php#L233%7CL234

The multiblog constant is simply displaying pages..There's 1 unique root blog that's why to me it's not needed to display this screen for other blogs that are relying on the unique BP instance.

That being said I may have missed a specific workaround making it possible to have multiple BP instances on a network. Can you light my ignorance about it?

src/bp-core/admin/bp-core-admin-tools.php
imath: Same reply

src/bp-core/admin/bp-core-admin-tools.php
renatonascalves: So, my recollection of that feature is wrong, then.

https://github.com/buddypress/buddypress/blob/41d7e95513c7c1441fc94e254333be2cb9abb80e/src/class-buddypress.php#L423
https://codex.buddypress.org/getting-started/customizing/bp_enable_multiblog/
https://github.com/buddypress/buddypress/blob/41d7e95513c7c1441fc94e254333be2cb9abb80e/src/bp-core/bp-core-functions.php#L2205

This here makes me think it is possible to have BP in a MU network but being the root of the site in the network.

https://github.com/buddypress/buddypress/blob/41d7e95513c7c1441fc94e254333be2cb9abb80e/src/readme.txt#L87-L90

Could we discuss removing this feature from BP? It seems pretty old, pre-1.7 version.

imath: I agree the multiblog mode is confusing. It doesn't mean you can have multiple BP instances. It means you can display the content of the same instance on multiple blogs. The readme text, you referenced is explaining it differently talking about "central data".

> to allow your BuddyPress content to be displayed on any site in your WordPress Multisite network, using the same central data.

I agree we should probably rethink all the possible configs plugin's history made available. Reading the codex page you added a link to seems to explain it can be interesting when you want to use it for translation purposes.

Maybe rethinking the Administration part we've included into our roadmap can help us simplify how BuddyPress can be configured.

renatonascalves: Agree! Thanks!!

imath:
Fixed in https://github.com/buddypress/buddypress/commit/ebf1d008f923b636cf3c2fecefb2651c31290757

PR #238, opened 2024-02-26
Introduce the Activity Type API and replace favorites with likes!

imath:
I'm back again with Favorites improvement. The story of this ticket highlighted the fact we should consider moving to **Likes**. I agree and now we have a Backward Compatibility Add-on (BP Classic). I believe we should move the Activity Favorites feature into this Add-on and finally have **Likes** instead into BuddyPress.

I've tested many ways to fix this ticket and I believe I finally have something solid to improve how we deal with **Likes** or more globally with **Reactions** to an activity. My reasoning to build the Activity Likes feature is Activity comments or Activity Likes are both reactions to an activity: I like what the activity says or I reply to it. So **Activity Likes** can be compared to Activity comments without content. If this PR is still a WIP, it already contains:

- An Activity Type API (it should help us fix [#6429](https://buddypress.trac.wordpress.org/ticket/6429)
- Examples of what improvements this API can bring: see the `bp_activity_type_supports()` function.
- An initial Reaction API I'm inaugurating with the **Activity Likes** feature.
- Adding/Removing Likes, getting likes count for an activity, getting the activity the user liked

Next steps if we agree on this scenario:
- List the users who liked an activity on the Activity single screen
- Ajax/JavaScript implementation into Template packs

More possible ambitious steps:
- Sort Activity stream according to number of activity likes
- Sort Activity stream according to number of comments.

I really think we should go this road for Activity Reactions/Likes. What do you think? Should we make this the 14.0.0 release top feature?

Trac ticket: https://buddypress.trac.wordpress.org/ticket/5644

---

### DIFF
--- a/src/bp-activity/actions/dislike.php
+++ b/src/bp-activity/actions/dislike.php
@@ -17,11 +17,11 @@ if ( ! defined( 'ABSPATH' ) ) {
  *
  * @since 14.0.0
  *
- * @return bool False on failure.
+ * @return void
  */
 function bp_activity_dislike_action() {
 	if ( ! is_user_logged_in() || ! bp_is_activity_component() || ! bp_is_current_action( 'dislike' ) ) {
-		return false;
+		return;
 	}
 
 	// Check the nonce.
--- a/src/bp-activity/actions/dislike.php
+++ b/src/bp-activity/actions/dislike.php
@@ -1,0 +1,69 @@
+<?php
+/**
+ * Activity: Dislike action
+ *
+ * @package BuddyPress
+ * @subpackage ActivityActions
+ * @since 14.0.0
+ */
+
+// Exit if accessed directly.
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Dislike an activity.
+ *
+ * @since 14.0.0
+ *
+ * @return bool False on failure.
+ */
+function bp_activity_dislike_action() {
+	if ( ! is_user_logged_in() || ! bp_is_activity_component() || ! bp_is_current_action( 'dislike' ) ) {
+		return false;
+	}
+
+	// Check the nonce.
+	check_admin_referer( 'bp_activity_like' );
+
+	$feedback        = __( 'Ouch something went wrong when trying to dislike this activity.', 'buddypress' );
+	$feedback_type   = 'error';
+	$activities      = bp_activity_get_specific(
+		array(
+			'activity_ids'     => bp_action_variable( 0 ),
+			'display_comments' => 'threaded',
+		)
+	);
+	$parent_activity = reset( $activities['activities'] );
+
+	if ( empty( $parent_activity->id ) ) {
+		$feedback = __( 'The activity you want to dislike does not exist.', 'buddypress' );
+
+	} else {
+		$user_like = array();
+		if ( isset( $parent_activity->reactions ) ) {
+			$user_like = wp_filter_object_list( $parent_activity->reactions, array( 'type' => 'activity_like', 'user_id' => bp_loggedin_user_id() ), 'AND', 'id' );
+		}
+
+		// The user didn't like it!
+		if ( empty( $user_like ) || 1 !== count( $user_like ) ) {
+			$feedback = __( 'Ouch! Are you sure you liked this activity? Looks like it’s not the case.', 'buddypress' );
+
+		} else {
+			$reaction_id = reset( $user_like );
+			$disliked    = bp_activity_remove_reaction( $reaction_id );
+
+			if ( is_wp_error( $disliked ) ) {
+				$feedback = $disliked->get_error_message();
+			} else {
+				$feedback      = __( 'You successfully disliked it!', 'buddypress' );
+				$feedback_type = 'success';
+			}
+		}
+	}
+
+	bp_core_add_message( $feedback, $feedback_type );
+	bp_core_redirect( wp_get_referer() );
+}
+add_action( 'bp_actions', 'bp_activity_dislike_action' );
--- a/src/bp-activity/bp-activity-filters.php
+++ b/src/bp-activity/bp-activity-filters.php
@@ -719,7 +719,7 @@ add_filter( 'bp_activity_set_just-me_scope_args', 'bp_activity_filter_just_me_sc
  *
  * @param array $retval Empty array by default.
  * @param array $filter Current activity arguments.
- * @return array $retval
+ * @return array
  */
 function bp_activity_filter_likes_scope( $retval = array(), $filter = array() ) {
 
--- a/src/bp-activity/bp-activity-filters.php
+++ b/src/bp-activity/bp-activity-filters.php
@@ -713,6 +713,62 @@ function bp_activity_filter_just_me_scope( $retval = array(), $filter = array() 
 add_filter( 'bp_activity_set_just-me_scope_args', 'bp_activity_filter_just_me_scope', 10, 2 );
 
 /**
+ * Set up activity arguments for use with the 'likes' scope.
+ *
+ * @since 14.0.0
+ *
+ * @param array $retval Empty array by default.
+ * @param array $filter Current activity arguments.
+ * @return array $retval
+ */
+function bp_activity_filter_likes_scope( $retval = array(), $filter = array() ) {
+
+	// Determine the user_id.
+	if ( ! empty( $filter['user_id'] ) ) {
+		$user_id = $filter['user_id'];
+	} else {
+		$user_id = bp_displayed_user_id()
+			? bp_displayed_user_id()
+			: bp_loggedin_user_id();
+	}
+
+	// Get user likes.
+	$likes = bp_activity_get_user_reactions( $user_id );
+	if ( empty( $likes ) ) {
+		$likes = array( 0 );
+	}
+
+	// Should we show all items regardless of sitewide visibility?
+	$show_hidden = array();
+	if ( ! empty( $user_id ) && ( $user_id !== bp_loggedin_user_id() ) ) {
+		$show_hidden = array(
+			'column' => 'hide_sitewide',
+			'value'  => 0
+		);
+	}
+
+	$retval = array(
+		'relation' => 'AND',
+		array(
+			'column'  => 'id',
+			'compare' => 'IN',
+			'value'   => (array) $likes,
+		),
+		$show_hidden,
+
+		// Overrides.
+		'override' => array(
+			'display_comments' => true,
+			'filter'           => array( 'user_id' => 0 ),
+			'show_hidden'      => true
+		),
+	);
+
+	return $retval;
+}
+add_filter( 'bp_activity_set_likes_scope_args', 'bp_activity_filter_likes_scope', 10, 2 );
+
+/**
  * Set up activity arguments for use with the 'favorites' scope.
  *
  * @since 2.2.0
--- a/src/bp-activity/bp-activity-functions.php
+++ b/src/bp-activity/bp-activity-functions.php
@@ -1217,11 +1217,6 @@ function bp_activity_add_reaction( $args = '' ) {
 		)
 	);
 
-	// Bail on failure.
-	if ( is_wp_error( $reaction_id ) ) {
-		return $reaction_id;
-	}
-
 	return $reaction_id;
 }
 
--- a/src/bp-activity/bp-activity-functions.php
+++ b/src/bp-activity/bp-activity-functions.php
@@ -1143,7 +1143,7 @@ function bp_activity_get_actions_for_context( $context = '' ) {
  *     @type bool       $skip_notification Optional. false to send a reaction notification, true otherwise.
  *                                         Defaults to true.
  * }
- * @return WP_Error|int The ID of the reaction on success, otherwise false.
+ * @return WP_Error|integer The ID of the reaction on success, otherwise false.
  */
 function bp_activity_add_reaction( $args = '' ) {
 	$r = bp_parse_args(
--- a/src/bp-activity/bp-activity-functions.php
+++ b/src/bp-activity/bp-activity-functions.php
@@ -375,18 +375,36 @@ function bp_register_activity_type( $type, $args = array() ) {
  *                                   WP_Error object on failure.
  */
 function bp_unregister_activity_type( $type ) {
+	$type_object = bp_get_activity_type_object( $type );
+
+	if ( is_wp_error( $type_object ) ) {
+		return $type_object;
+	}
+
+	unset( buddypress()->activity->types[ $type_object->name ] );
+
+	return true;
+}
+
+/**
+ * Retrieves the list of Activity types having a specific role.
+ *
+ * @since 14.0.0
+ *
+ * @param string $type The name key of the activity type.
+ * @return WP_Error|BP_Activity_Type The list of Activity type name keys having the requested role.
+ */
+function bp_get_activity_type_object( $type ) {
 	$activity_types = buddypress()->activity->types;
 
-	// Sanitize post type name.
+	// Sanitize activity type name.
 	$type = sanitize_key( $type );
 
 	if ( ! isset( $activity_types[ $type ] ) ) {
 		return new WP_Error( 'invalid_activity_type', __( 'Invalid activity type.', 'buddypress' ) );
 	}
 
-	unset( buddypress()->activity->types[ $type ] );
-
-	return true;
+	return $activity_types[ $type ];
 }
 
 /**
@@ -792,6 +810,7 @@ function bp_activity_type_supports( $activity_type = '', $feature = '' ) {
 	switch ( $feature ) {
 		case 'comments' :
 		case 'favorites' :
+		case 'likes' :
 			$retval = isset( $bp->activity->types[ $activity_type ]->supports->{$feature} ) && true === $bp->activity->types[ $activity_type ]->supports->{$feature};
 			break;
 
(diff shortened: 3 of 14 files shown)


emaralive:
Maybe if I explain it this way, you may gain a different perspective. Currently, I am a member of a site that has ability to **Like** activity posts in addition to being able to **Favorite** activity posts. As a matter of fact, the **site admin** has relabeled **Favorites** to read as **Bookmark** which allows members to have a list of activity posts that they can just refer to because:

- They can **Like** the post at a later time/date.
- They can **Comment** on the post at a later time/date.
- They can later refer to the post for sharing purposes at a later time/date.

Seems to me, you are creating an either this or either that situation by moving the **Favorites** functionality into **BP Classic**.  I'm not sure why **site admins** can't have both as is currently possible now.

src/bp-activity/actions/dislike.php
```diff
@@ -0,0 +1,69 @@
...
+
+	// Check the nonce.
+	check_admin_referer( 'bp_activity_like' );
+
+	$feedback        = __( 'Ouch something went wrong when trying to dislike this activity.', 'buddypress' );
+	$feedback_type   = 'error';
+	$activities      = bp_activity_get_specific(
+		array(
+			'activity_ids'     => bp_action_variable( 0 ),
+			'display_comments' => 'threaded',
```
renatonascalves, line 35: Why is this necessary?

src/bp-activity/actions/dislike.php
```diff
@@ -0,0 +1,69 @@
...
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
+
+/**
+ * Dislike an activity.
+ *
+ * @since 14.0.0
+ *
+ * @return bool False on failure.
```
renatonascalves, line 20: This has a mixture of `bool` with `void`. Tools like `phpstan` would rightly complain it is missing a return value. I think you should bail without a `bool`.

src/bp-activity/actions/dislike.php
```diff
@@ -0,0 +1,69 @@
...
+ */
+function bp_activity_dislike_action() {
+	if ( ! is_user_logged_in() || ! bp_is_activity_component() || ! bp_is_current_action( 'dislike' ) ) {
+		return false;
```
renatonascalves, line 24: See my comment above.

```suggestion
		return;
```

renatonascalves: Since `add_action` doesn't expect a returned value, bailing into `void` is better here.

src/bp-activity/bp-activity-filters.php
```diff
@@ -712,6 +712,62 @@ function bp_activity_filter_just_me_scope( $retval = array(), $filter = array()
...
 add_filter( 'bp_activity_set_just-me_scope_args', 'bp_activity_filter_just_me_scope', 10, 2 );
 
+/**
+ * Set up activity arguments for use with the 'likes' scope.
+ *
+ * @since 14.0.0
+ *
+ * @param array $retval Empty array by default.
+ * @param array $filter Current activity arguments.
+ * @return array $retval
```
renatonascalves, line 722:
```suggestion
 * @return array
```

src/bp-activity/bp-activity-functions.php
```diff
@@ -311,7 +311,114 @@ function bp_activity_get_userid_from_mentionname( $mentionname ) {
...
+ *     @type array    $supports        The list of supported features for the activity type.
+ *                                     Default [].
+ * }
+ * @return BP_Activity_Type|WP_Error The registered activity type object on success,
+ *                                   WP_Error object on failure.
+ */
+function bp_register_activity_type( $type, $args = array() ) {
+	$activity_types = buddypress()->activity->types;
+
+	// Sanitize post type name.
```
renatonascalves, line 350: Not a post type.

src/bp-activity/bp-activity-functions.php
```diff
@@ -311,7 +311,114 @@ function bp_activity_get_userid_from_mentionname( $mentionname ) {
...
+function bp_unregister_activity_type( $type ) {
+	$type_object = bp_get_activity_type_object( $type );
+
+	if ( is_wp_error( $type_object ) ) {
+		return $type_object;
+	}
+
+	unset( buddypress()->activity->types[ $type_object->name ] );
+
+	return true;
```
renatonascalves, line 386: > The registered activity type object on success

is this right?

src/bp-activity/bp-activity-functions.php
```diff
@@ -311,7 +311,114 @@ function bp_activity_get_userid_from_mentionname( $mentionname ) {
...
+/**
+ * Retrieves the list of Activity types having a specific role.
+ *
+ * @since 14.0.0
+ *
+ * @param string $role The requested role for activity types to retrieve.
+ * @return array The list of Activity type name keys having the requested role.
+ */
+function bp_get_activity_types_for_role( $role ) {
+	$types = wp_filter_object_list( buddypress()->activity->types, array( 'role' => $role ), 'and', 'name' );
```
renatonascalves, line 419: Do we have a get types sort of function? I see we are always getting it from the bp global...

src/bp-activity/bp-activity-functions.php
```diff
@@ -1011,6 +1123,226 @@ function bp_activity_get_actions_for_context( $context = '' ) {
...
+/**
+ * React to an activity.
+ *
+ * @since 14.0.0
+ *
+ * @param array|string $args {
+ *     An array of arguments.
+ *     @type int        $user_id           Optional. The ID of the user reacting.
+ *                                         Defaults to the ID of the logged-in user.
+ *     @type object|int $activity          Required. The parent activity object or its ID.
```
renatonascalves, line 1137: Better to say what object.
```suggestion
 *     @type BP_Activity_Activity |int $activity          Required. The parent activity object or its ID.
```

src/bp-activity/bp-activity-functions.php
```diff
@@ -1737,6 +2148,33 @@ function bp_activity_format_activity_action_activity_comment( $action, $activity
...
 }
 
+/**
+ * Format 'activity_like' activity actions.
+ *
+ * @since 14.0.0
+ *
+ * @param string $action   Static activity action.
+ * @param object $activity Activity data object.
+ * @return string $action
```
renatonascalves, line 2158:
```suggestion
 * @return string
```

src/bp-activity/bp-activity-functions.php
```diff
@@ -1737,6 +2148,33 @@ function bp_activity_format_activity_action_activity_comment( $action, $activity
 	return apply_filters( 'bp_activity_comment_action', $action, $activity );
 }
 
+/**
+ * Format 'activity_like' activity actions.
+ *
+ * @since 14.0.0
+ *
+ * @param string $action   Static activity action.
+ * @param object $activity Activity data object.
```
renatonascalves, line 2157:
```suggestion
 * @param BP_Activity_Activity $activity Activity data object.
```

src/bp-activity/bp-activity-template.php
```diff
@@ -4322,3 +4327,142 @@ function bp_get_activity_show_filters( $context = '' ) {
...
+
+	if ( $activity_id ) {
+		$user_like = bp_activity_get(
+			array(
+				'type'             => 'activity_like',
+				'filter'           => array(
+					'user_id'    => $user_id,
+					'primary_id' => $activity_id,
+				),
+				'display_comments' => 'stream',
```
renatonascalves, line 4372: Do we need to append the comments too? This could be expensive.

src/bp-activity/bp-activity-template.php
```diff
@@ -4322,3 +4327,142 @@ function bp_get_activity_show_filters( $context = '' ) {
...
+	/**
+	 * Filters the link to like an activity.
+	 *
+	 * @since 14.0.0
+	 *
+	 * @param string $url  Constructed link for liking the activity.
+	 * @param string $type Whether to `add` or `remove` the like.
+	 * @param string $slug The computed slug according to the type (`like` or `dislike`).
+	 */
+	return apply_filters( 'bp_get_activity_like_link', $url, $type, $slug );
```
renatonascalves, line 4467: Suggestion: pass the activity object too.

src/bp-activity/classes/class-bp-activity-component.php
```diff
@@ -19,6 +19,13 @@
  */
 #[AllowDynamicProperties]
 class BP_Activity_Component extends BP_Component {
+	/**
+	 * Types.
+	 *
+	 * @since 14.0.0
+	 * @var array $types
```
renatonascalves, line 26:
```suggestion
	 * @var array
```

src/bp-activity/classes/class-bp-activity-activity.php
```diff
@@ -1573,38 +1610,299 @@ public static function delete_activity_meta_entries( $activity_ids = array() ) {
...
 		return $activities;
 	}
 
+	/**
+	 * Append activity comments to their associated activity items.
+	 *
+	 * @since 1.2.0
+	 * @deprecated 14.0.0
+	 *
+	 * @global wpdb $wpdb WordPress database object.
```
renatonascalves, line 1889: This can be removed.

renatonascalves (changes requested):
I agree with @emaralive

Let's leave to the community owner to decide how they wanna use each feature. See Instagram for an example of a community where people can like an activity, but you can also bookmark/favorite a post/activity. Each serves different purposes.

I also think this is being attached to the Activity component, but a plugin developer could very well build a feature to like groups, messages, users, or posts. The BuddyPress Followers plugin has a nice API that allows for extension for different scenarios. Something we could use as an inspiration here.

I left a few notes to the code. :)

imath:
@emaralive & @renatonascalves 

Thanks a lot for your feedbacks, I'm just following the consensus that was found about choosing both features, favorites or likes only:
See https://buddypress.trac.wordpress.org/ticket/5644#comment:36

Moreover, Favorites are very buggy, we really should deprecate it. If there's a strong willing to absolutely keep the idea of Bookmarking I'll surrender but in this case it should be rebuilt as the Like feature using an activity type having a `reaction` role.

renatonascalves:
@imath I don't have a strong opinion about it. To be honest, I'm not sure if this feature is used a lot in communities.

If we can deprecate it without any major issues to users, great. I do have a strong opinion that the API should be flexible enough that plugin developers can use it for other components, if desired.

BuddyPress is very good at being extendable to developers. The current API in this pr could do with some improvements, getting the reactions is a bit burdensome, lots of PHP filtering.

src/bp-activity/actions/dislike.php
imath: To validate the "like" exists, we'll need to deprecated `display_comments` in favor of `display_reactions`. This is necessary because likes as comments are Activity children.

imath:
@renatonascalves thanks for your review by the way, I'll update the PR accordingly. This is still a WIP but I believe the Activity Type API combined with the reaction one will improve & simplify things.

emaralive:
> @emaralive & @renatonascalves
> 
> Thanks a lot for your feedbacks, I'm just following the consensus that was found about choosing both features, favorites or likes only: See https://buddypress.trac.wordpress.org/ticket/5644#comment:36
> 
@imath & @renatonascalves 

Seems to me the referenced comment was from **9 years ago**, of which, that was the thought then (2015) and this is now (2024).  AFAIK, things/thoughts/conditions change over time. 

 > Moreover, Favorites are very buggy, we really should deprecate it. If there's a strong willing to absolutely keep the idea of Bookmarking I'll surrender but in this case it should be rebuilt as the Like feature using an activity type having a `reaction` role.

As to buggy, could you expand upon this notion, @imath? The last bug I remember is reconciling Activity IDs after an Activity/Comment was deleted.

I'm trying to imagine how **Slack** would work without the ability to **Save for later** (bookmark); meaning, I can only react to a message. I suppose I could manually copy the message link and save it in a text file but, that would actually be archaic and I shudder at the thought.

However, in all fairness, I've not checked out this PR to see how it actually works and when I get an opportunity, I will do so.  If, by chance, there isn't the ability to **bookmark** then I would recommend the exploration of "_it should be rebuilt as the Like feature using an activity type having a `reaction` role_" because I don''t know what that means nor what it would look like.

dcavins:
I agree that bookmarking/favoriting and liking posts are two different types of actions, and while adding reactions is a great improvement (and probably expected funtionality for most users), I'd hate to lose bookmarks, for those users who are bookmarkers. If bookmarks are buggy, we should fix them. I'm guessing that one of the issues is that the old code stored them in a serialized array (I'm guessing about this), so when an item is deleted, it's hard to remove them from all of the serialized arrays. 

Reactions are great!

imath:
Ahaha thanks a lot 😍 for your reactions 😉 @dcavins & @emaralive.

As David explained using serialized information into a user meta will always be a trouble and we need to avoid doing this as much as we can (we sometimes have no other choices).

Favorites are already gone, reading your 2 comments: you're now talking about bookmarks so I think we can satisfy everyone's expectations:

1. the current "Favorites" is "classic" BuddyPress and needs to be deprecated, moved to BP Classic so that very conservative users can always enjoy it.
2. Modern BuddyPress is "Likes" & "Bookmarks".

**Likes**: will help make the most interesting activities to rise once we'll be able to sort activities according to the number of likes or comments. So that's the most interesting feature to me, it's a community feature in the sense that thanks to the public expression of the community members I can see the best of activities.

**Bookmarks** are a "selfish" need, it's less interesting to me, but, I was actually using it on Twitter 😁 (I'm not saying the before last letter of the Alphabet, because I left Twitter before it was renamed). So "technically" speaking: to me, "Bookmarks" are hidden "likes" only visible to me. I can go to my list of bookmarks, read the activity, then remove the bookmark once I've read it. I agree it's interesting to build this feature using reactions and probably use this feature to eventually migrate the favorited activities that are still existing.

But Favorites are different, it's more persistent. My favorite movie is Matrix, my favorite pizza is Pepperoni : it won't change... I think BP Favorites were used as bookmarks.

In short can we agree on, let's build 2 Activity reactions in 14.0.0:
- Likes
- Bookmarks

and keep Favorites, but, into the BP Classic Add-on 🧐 ?

renatonascalves:
Technically speaking, will I be able to add Group reactions? Like a group? Bookmark a group? A blog post? Like a message? Bookmark a message (similar to Slack)?

Let's not build the API attached to the Activity component since that's not the only one a site might need it.  So naming Activity reactions might be the wrong start. Having a "Reactions API" might be better.

imath:
💥

Thanks @renatonascalves for your feedback. I understand you expect something larger than Activity "Reactions". Maybe the "Reaction" word is a problem, btw. To me a comment is already an activity reaction, for example. 

1) I'm afraid what you seem to suggest won't be achieved in 4 months for 14.0.0. A centralized Reaction API makes me think about this project that never happened: https://github.com/buddypress/bp-relationships
2) Thanks for insisting about https://github.com/r-a-y/buddypress-followers I wasn't aware it included features to subscribe/follow an activity or a blog. I thought it was only targeting users.
3) FWIW: about liking, bookmarking a group, a post, a blog : you'd be able to do so liking or bookmarking the activity generated by these objects. To me liking a group or a blog does less sense than an activity, a comment or a post by the way.
4) Messages: the star feature is doing a nice job, I believe.

I'm not as ambitious as you seem to be. I'm suggesting to satisfy a need I often read on BuddyPress support or in reviews about it, find a way to:
- introduce Activity likes (as a replacement of favorites was my understanding of the ticket's history).
- I thought it could replace gracefully Favorites, but understood the majority of the team thinks we should keep this feature
- introducing Activity bookmarks using some kind of hidden likes can be done for 14.0.0 and I hope can satisfy @emaralive & @dcavins concerns.

But this ticket is not about building a central Reaction API (I'll edit the PR's title as I admit it was misleading), it's about improving favorites management. I'm now trying another alternative way using activity children to create likes or bookmarks just like we do for comments.

So:
- If the "Reaction" word is a problem, I'm fine with using another one 😃.
- If using activity children to manage likes or bookmarks is a problem or everyone thinks a central reaction API is the right way to fix this ticket, I can halt there and move to another ticket.

I begin to believe this ticket is cursed 😂 🧙‍♀️ At least, I've tried many ways to improve the current Favorites situation for 10 years!

renatonascalves:
Agree! Let's move the convo for an extendable API for https://github.com/buddypress/bp-relationships or a related ticket. I mistakenly assumed that was the goal here.

> Thanks for insisting about https://github.com/r-a-y/buddypress-followers I wasn't aware it included features to subscribe/follow an activity or a blog. I thought it was only targeting users.

I used it before to add a feature to a client site. It has a good database schema that made the job a breeze.

src/bp-activity/bp-activity-functions.php
imath: We already have `bp_activity_get_types()` See https://github.com/buddypress/buddypress/blob/54d3c6140bf0a773f06032f8bf9eb112404d3fe7/src/bp-activity/bp-activity-functions.php#L886

src/bp-activity/bp-activity-functions.php
imath: Problem is the function to get the activity to add a like for into the corresponding `src/bp-activity/actions/like.php` file is `bp_activity_get_specific()` which is not returning a `BP_Activity_Activity` object but a standard one. So I'll leave it this way for now.

src/bp-activity/bp-activity-functions.php
imath: As it is used to inside the function to fetch a list of activities being regular objects, I'll leave it this way for now. See https://github.com/buddypress/buddypress/blob/54d3c6140bf0a773f06032f8bf9eb112404d3fe7/src/bp-activity/classes/class-bp-activity-activity.php#L836

src/bp-activity/bp-activity-template.php
imath: `display_comments` should be renamed so that it works for other type of activity reactions. This is required because `activity_like` is always attached to a parent activity, so you need to get activity children: this is only happening when `'display_comments' => 'stream'` in BuddyPress .

imath: Btw, a bit above, there's a type filter so that only `activity_like` one are returned + this type does not support comments.

src/bp-activity/classes/class-bp-activity-activity.php
imath: Ok, good to know. I'll leave it, just like WordPress does.

imath:
I just refreshed the PR, but I won't work on improving the Activity Favorites / Likes (or eventually Activity Bookmarks) feature anymore (although "Likes" is one of the features users expect to finally get in BuddyPress just like it was the case for the private visibility).

I understood you all disagree with my strong conviction that **it has to be managed using Activity children** just like Activity Comments.

Let's wait for an hypothetical more global BP Reactions API to land one of these days in BuddyPress (eg: https://github.com/buddypress/bp-relationships)... The ironical part about it would probably be that users will want these reactions to generate activity entries (duplicate info we could have avoided with the way I'm convinced we should use). Nevermind.

Instead I'll pick the part about the Activity Type "API" to clean the mess currently existing between Activity Actions & Activity Types from this ticket [#6429](https://buddypress.trac.wordpress.org/ticket/6429).

renatonascalves:
Just adding my two cents here that, I know you are unhappy with this outcome, and I totally understand the frustration.

However, long term, I do think having a relationship API is the right approach to take, even if it takes longer to get there. I'm personally invested in this feature, so you can be sure we will make some progres in the next release.

> The ironical part about it would probably be that users will want these reactions to generate activity entries (duplicate info we could have avoided with the way I'm convinced we should use).

^ here I think we will need to have better integration not only with the activity component, but also with the notifications component, the settings component (so that one can turn off notifications), etc.

There is another ticket related to the notifications part somewhere. But the pieces are all connected. Your work on the Activity Type "API" will help with this task too. :)

PR #271, opened 2024-04-14
Deprecate the backcompat feature about creating a user out of a signup

imath:
This backwards compatibility feature was introduced in version 2.0.0, it's causing an issue as the just registered user's profile can already be seen on the front-end. Moreover as it's not yet a user, let's be consistent and only deal with our BP Signups API to manage signups.

The goal of this PR is to stop creating a user on regular WordPress site when a registration is performed.

Trac ticket: https://buddypress.trac.wordpress.org/ticket/6123

---

### DIFF
--- a/src/bp-members/bp-members-functions.php
+++ b/src/bp-members/bp-members-functions.php
@@ -1860,19 +1860,15 @@ function bp_core_signup_user( $user_login, $user_password, $user_email, $usermet
 		$activation_key = wp_generate_password( 32, false );
 
 		/**
-		 * WordPress's default behavior is to create user accounts
-		 * immediately at registration time. BuddyPress uses a system
-		 * borrowed from WordPress Multisite, where signups are stored
-		 * separately and accounts are only created at the time of
-		 * activation. For backward compatibility with plugins that may
-		 * be anticipating WP's default behavior, BP silently creates
-		 * accounts for registrations (though it does not use them). If
-		 * you know that you are not running any plugins dependent on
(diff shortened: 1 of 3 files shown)


src/bp-members/bp-members-functions.php
```diff
@@ -1860,19 +1860,15 @@ function bp_core_signup_user( $user_login, $user_password, $user_email, $usermet
...
-		 * to true in your wp-config.php file.
+		 * Filter here to keep creating a user when a registration is performed on regular WordPress configs.
+		 *
+		 * @since 14.0.0
+		 * @todo Fully deprecate in 15.0.0
+		 *
+		 * @param boolean $value True to carry on creating a user when a registration is performed.
+		 *                       False otherwise.
 		 */
-		if ( ! defined( 'BP_SIGNUPS_SKIP_USER_CREATION' ) || ! BP_SIGNUPS_SKIP_USER_CREATION ) {
```
renatonascalves, line 1875: For those using this constant, I wonder if we should add a `_doing_it_wrong`. Similar to how we are doing with other similar constants.

renatonascalves: The default value of this hook could be the value of the `BP_SIGNUPS_SKIP_USER_CREATION` constant itself. In case someone relies on it.

renatonascalves (approved):

imath:
Thanks for your suggestions here and your feedback on the Trac ticket @renatonascalves I will update the PR asap.

PR #273, opened 2024-04-19
Unit tests for the #8728 ticket

renatonascalves:
Trac ticket: https://buddypress.trac.wordpress.org/ticket/8728

---
**This Pull Request is for code review only. Please keep all other discussion in the BuddyPress Trac ticket. Do not merge this Pull Request. See [GitHub Pull Requests for Code Review](https://make.wordpress.org/core/handbook/contribute/git/github-pull-requests-for-code-review/) in the WordPress Core Handbook for more details.**

PR #278, opened 2024-05-02
Apply @emaralive patch, copy @espellcaste unit test & improve some parts

imath:
+ Include tests from this PR https://github.com/buddypress/buddypress/pull/273 (putting them into a more appropriate file).
+ Add some other tests to make sure a group admin cannot delete any activity, an activity posted in a group he's not a member of, or activities of the group he's an admin of when the Admin globally disabled the group activity deletion.
+ Use the same logic than other options:  by default group Admins/Mods can delete activities. The Admin needs to uncheck the global settings to disable it.
+ Let's rename options/functions the "disable" way !

PS: @emaralive, @needle is right (thanks a lot for your input), so I've updated the patch accordingly.

Trac ticket: https://buddypress.trac.wordpress.org/ticket/8728

---

### DIFF
--- a/src/bp-core/admin/bp-core-admin-settings.php
+++ b/src/bp-core/admin/bp-core-admin-settings.php
@@ -351,6 +351,18 @@ function bp_admin_setting_callback_group_cover_image_uploads() {
 <?php
 }
 
+/**
+ * 'Enable group activity deletions.
+ *
+ * @since 14.0.0
+ */
+function bp_admin_setting_callback_group_activity_deletions() {
+?>
+	<input id="bp-disable-group-activity-deletions" name="bp-disable-group-activity-deletions" type="checkbox" value="1" <?php checked( ! bp_disable_group_activity_deletions() ); ?> />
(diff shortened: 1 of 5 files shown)


emaralive:
@imath, I appear to be experiencing unexpected behavior with the way the **Group Activity Deletions** checkbox is working. IOW, by default, the  checkbox is now checked (selected), I would expect that deselecting the checkbox and then saving the change (save settings)  would result with a deselected checkbox. 

However, the checkbox remains checked (selected) or reverts back to being checked (selected). Can you confirm or deny this behavior?

imath:
Ouch! Thanks for your alert about it @emaralive I'll give the PR more testing to see if I missed something 👌

imath:
@emaralive just updated the PR, the option should be saved now, can you confirm?

emaralive:
@imath I just finished testing the **Group Activity Deletions** checkbox and can confirm that the checkbox now behaves as expected.

PR #285, opened 2024-05-05
Use the right user ID to accept/reject group invites

imath:
Unless an Administrator is viewing another member's group invitations screen, the right user ID to use is the logged in user ID.

Trac ticket: https://buddypress.trac.wordpress.org/ticket/9135

---


dcavins:
Your fix looks good to me! I agree that it should nearly always be the logged-in user's ID, with the very unusual exception that an admin is helping another user accept an invite, which you've covered.

PR #287, opened 2024-05-09
Playground compat: stop using `bool` db field type

imath:
Updates BuddyPress db tables schema so that it uses tinyint fields instead of boolean ones. This is required to have WP Playground work as expected.

NB: this move is requiring a major version to be achieved.

Trac ticket: https://buddypress.trac.wordpress.org/ticket/9126

---


blueprint.json
```diff
@@ -0,0 +1,18 @@
+{
+	"$schema": "https://playground.wordpress.net/blueprint-schema.json",
+	"landingPage": "\/wp-admin\/admin.php?page=bp-components",
+	"preferredVersions": {
+		"php": "8.1",
```
renatonascalves, line 5: Not PHP 8.2 or PHP 8.3?

blueprint.json
```diff
@@ -0,0 +1,18 @@
...
+		"wp": "6.5"
```
renatonascalves, line 6: Is it possible to set it to `latest`?

blueprint.json
imath: Good point!
```suggestion
		"php": "8.2",
```

blueprint.json
imath: It is! Just checked: https://wordpress.github.io/wordpress-playground/blueprints-api/index/
```suggestion
		"wp": "latest"
```

PR #293, opened 2024-05-21
Use better hooks to notify group members they were promoted

imath:
Hi @needle

Just checked, I believe we actually need to update BuddyPress and the BP REST API.

- BuddyPress: it shouldn't use the current hooks as there can be situations where the hook is fired although the requested action failed.
- BuddyPress `groups_$action_member()` function should be usable outside of the Web version using the BP REST API for instance. It's not currently the case and I suspect it to be the reason why we were directly using `BP_Groups_Member->$action()`
- The BP REST API should absolutely use `groups_$action_member()` functions to be sure cache is cleared or group activities are deleted when it's needed. It's currently not the case. I'll soon submit a PR to the BP REST API GH repo.

Trac ticket: https://buddypress.trac.wordpress.org/ticket/9158

---

### DIFF
--- a/src/bp-groups/bp-groups-functions.php
+++ b/src/bp-groups/bp-groups-functions.php
@@ -2006,21 +2006,32 @@ function groups_delete_all_group_invites( $group_id ) {
  * Promote a member to a new status within a group.
  *
  * @since 1.0.0
+ * @since 14.0.0 Adds the `$group_admin_id` parameter.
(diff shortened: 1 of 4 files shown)


imath:
Fixed in https://github.com/buddypress/buddypress/commit/e667c79966b84ee9e0d230a8bac7bddf6b53926b

PR #310, opened 2024-06-02
Corrected @since Order in Php Docs

shail-mehta:
Trac Ticket: [9168](https://buddypress.trac.wordpress.org/ticket/9168)

### DIFF
--- a/src/bp-core/admin/bp-core-admin-functions.php
+++ b/src/bp-core/admin/bp-core-admin-functions.php
@@ -26,11 +26,11 @@ function bp_core_admin_menu_init() {
  *
  * The old "bp-general-settings" page was renamed "bp-components".
  *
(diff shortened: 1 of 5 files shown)


src/bp-core/admin/bp-core-admin-functions.php
```diff
@@ -220,9 +220,9 @@ function bp_core_add_admin_notice( $notice = '', $type = 'updated' ) {
  *   - that no WP page has multiple BP components associated with it.
  * The administrator will be shown a notice for each check that fails.
  *
- * @global wpdb $wpdb WordPress database object.
```
renatonascalves, line 223: oh, had no idea it needed to be after. Goog catch.

renatonascalves (commented):
all looks good here.

PR #316, opened 2024-06-16
Bring back custom translation files for BuddyPress!

imath:
Use the `'load_translation_file'` filter to check custom file locations for the `buddypress` text domain.

**NB**: this filter was introduced by WordPress in version 6.5.

If there's a `buddypress-xx_XX.mo` (where xx_XX is the locale code) file into one of the following directories, it will be used to replace the default one (WordPress.org Plugin Directory / GlotPress):

1. `/wp-content/languages/plugins/buddypress` (added in 14.0.0)
2. `/wp-content/languages/buddypress`
3. `/wp-content/languages`

Trac ticket: https://buddypress.trac.wordpress.org/ticket/9187

---

### DIFF
--- a/src/bp-core/bp-core-functions.php
+++ b/src/bp-core/bp-core-functions.php
@@ -2484,6 +2484,54 @@ function bp_is_get_request() {
 /** Miscellaneous hooks *******************************************************/
 
 /**
+ * Looks for the requested file name into a list of custom language locations.
+ *
+ * @since 14.0.0
+ *
+ * @param string $file_name The file name.
+ * @return string A file path or an empty string if no files were found into custom language locations.
+ */
+function bp_get_custom_translation_file( $file_name = '' ) {
+	$file_path = '';
+
+	if ( $file_name ) {
+		/**
+		 * Filters the locations to load language files from.
+		 *
+		 * Custom translation files can be put in:
+		 * 1. `/wp-content/languages/plugins/buddypress`
+		 * 2. `/wp-content/languages/buddypress`
+		 * 3. `/wp-content/languages`
+		 *
+		 * @since 2.2.0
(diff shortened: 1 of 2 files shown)


src/bp-core/bp-core-functions.php
```diff
@@ -2484,49 +2484,69 @@ function bp_is_get_request() {
...
  *
- * @see load_textdomain() for a description of return values.
- *
- * @return bool
+ * @param  string $file   Absolut path to the translation file to use.
+ * @param  string $domain The text domain to check against `buddypress`.
+ * @param  string $locale The current locale for the WordPress site.
+ * @return string Absolut path to the translation file to use.
  */
-function bp_core_load_buddypress_textdomain() {
```
renatonascalves, line 2495: Do we have any unit tests for the old implementation?

renatonascalves (approved):
🌵

src/bp-core/bp-core-functions.php
imath: We don't have unit tests for old implementation. I've been looking into it to add some, but `WP_LANG_DIR` in WordPress Tests Suite can't be overridden, which means we need to rely on the dummy plugin WordPress uses, fake a translation and replicate the fact one or more of our strings are returned before `bp_core_loaded` is fired, in this case, `_load_textdomain_just_in_time()` will always set the location to the GlotPress one and `bp_core_load_buddypress_textdomain()` is useless.

I believe adding a unit test about it is not necessary.

renatonascalves: Sounds good.

imath:
FYI, I've also added a function to override default JavaScript translation files used in BP Blocks.

PR #329, opened 2024-07-10
Access to `wp-signup.php` should be allowed when Sites component is not active

renatonascalves:
Trac ticket: https://buddypress.trac.wordpress.org/ticket/8775

---

### DIFF
--- a/src/bp-members/bp-members-functions.php
+++ b/src/bp-members/bp-members-functions.php
@@ -2552,6 +2552,12 @@ function bp_core_wpsignup_redirect() {
 	} elseif ( $referer ) {
 		$referer_path     = wp_parse_url( $referer, PHP_URL_PATH );
 		$is_site_creation = false !== strpos( $referer_path, 'wp-admin/my-sites.php' );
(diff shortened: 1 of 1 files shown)


src/bp-members/bp-members-functions.php
```diff
@@ -2552,6 +2552,12 @@ function bp_core_wpsignup_redirect() {
 	} elseif ( $referer ) {
 		$referer_path     = wp_parse_url( $referer, PHP_URL_PATH );
 		$is_site_creation = false !== strpos( $referer_path, 'wp-admin/my-sites.php' );
+	} else {
```
renatonascalves, line 2555: This odd behavior only happens if one tries to access the `https://wp.test/wp-signup.php` directly while being logged in. Logged out, one is redirected to `https://wp.test/register/`. 

The current behavior expects the `$referer` to be present, meaning that only those accessing from inside WordPress, specifically from here `https://wp.test/wp-admin/my-sites.php`, which is not always the case.

PR #337, opened 2024-07-15
BP REST API - V2

renatonascalves:
Trac ticket: https://buddypress.trac.wordpress.org/ticket/9145

---


src/bp-blogs/bp-blogs-template.php
```diff
@@ -373,13 +373,17 @@ function bp_get_blog_avatar( $args = '' ) {
 		// Set default value for the `alt` attribute.
 		$alt_attribute = __( 'Site icon for the blog', 'buddypress' );
 
-		if ( ! $blogs_template && isset( $args['blog_id'] ) && $args['blog_id'] ) {
+		$has_current_blog = isset( $blogs_template ) && isset( $blogs_template->blog->blog_id );
+
+		if ( false === $has_current_blog && isset( $args['blog_id'] ) && $args['blog_id'] ) {
```
renatonascalves, line 378: IMO, we should respect if the `blog_id` is passed, regardless if there is a blog in the current loop. But, I assume this logic is here for a reason, so let's keep it.

But technically, **it's not possible** to pass another, different `blog_id` into `bp_get_blog_avatar` inside a blog loop because of this logic.

src/bp-core/bp-core-rest-api.php
```diff
@@ -10,34 +10,6 @@
...
-/**
- * Is the BP REST plugin is active?
- *
- * @since 5.0.0
- *
- * @return bool True if the BP REST plugin is active. False otherwise.
- */
-function bp_rest_is_plugin_active() {
-	return (bool) has_action( 'bp_rest_api_init', 'bp_rest' );
-}
```
imath, line 22: There's no emergency to remove this imho: it could be useful if in the future we need to provide a way to load a previous version of the BP REST API.

src/bp-core/classes/class-bp-component.php
```diff
@@ -621,7 +622,7 @@ public function setup_actions() {
 		add_action( 'bp_generate_rewrite_rules', array( $this, 'generate_rewrite_rules' ), 10 );
 
 		// Register BP REST Endpoints.
-		if ( bp_rest_in_buddypress() && bp_rest_api_is_available() ) {
+		if ( bp_rest_api_is_available() ) {
```
imath, line 625:
```suggestion
		if ( bp_rest_api_is_available() && ! bp_rest_is_plugin_active() ) {
```

It could be useful to avoid loading the built in BP REST API  if in the future we need to provide a way to load a previous version of the BP REST API.

src/bp-core/deprecated/15.0.php
```diff
@@ -0,0 +1,46 @@
...
+ * @since 5.0.0
+ * @deprecated 15.0.0
+ *
+ * @return bool True if the BP REST plugin is active. False otherwise.
+ */
+function bp_rest_is_plugin_active() {
+	_deprecated_function( __FUNCTION__, '15.0.0' );
+
+	return (bool) has_action( 'bp_rest_api_init', 'bp_rest' );
+}
```
imath, line 46:
```suggestion
```

We should keep this function as it could be useful to avoid loading the built in BP REST API if in the future we need to provide a way to load a previous version of the BP REST API.

imath (changes requested):
Hi @espellcaste 

Thanks a lot for your work on this.

- I was very surprised by the inclusion of V1 into class names: what is the purpose of `bp_rest_version()` then?
- Shouldn't `wp-json/buddypress/v1` results in loading classes for the version 1 of the BP REST API, `wp-json/buddypress/v2` results in loading classes for the version 2 of the BP REST API, and so on?
- What happens to plugins using directly current BP REST API classes? If you change their prefix in favor of `BP_REST_V1`, I'm afraid we would end up with fatal errors.
- According to this [comment](https://github.com/WP-API/WP-API/issues/1119#issuecomment-94594892) about the [WP API feature as a plugin](http://wp-api.org/index-deprecated.html), we should keep `BP_REST_` class prefix for v1.

**I strongly advise you to keep class names the way they currently are**: do not use `BP_REST_V1` prefixes. We need to preserve Plugin backwards compatibility.

renatonascalves:
This is almost ready.

renatonascalves:
This pull request is ready to be tested. There is some outstanding [questions](https://wordpress.slack.com/archives/C02RQBYUG/p1724459227573469) for some behaviors.

I'd recommend, for now, to only look at the code/implementation and offer suggestions before diving into testing mode. Just so that by the time we go to QA it, we'll be at a more solid state.

src/bp-blogs/bp-blogs-template.php
imath: I agree with you, maybe we can fix it later in another ticket?

src/bp-core/bp-core-functions.php
```diff
@@ -4986,29 +4986,31 @@ function bp_get_deprecated_functions_versions() {
...
+
+						return null;
 					},
 					scandir( buddypress()->plugin_dir . 'bp-core/deprecated' )
 				)
 			);
 
 			if ( array_diff( $deprecated_files, $deprecated_functions_versions ) ) {
-				return false;
+				return array();
```
imath, line 5013: Please don't change this, it's used to check if we forgot to add a deprecated version into the array. Moreover, it's not about your PR's goal. If you absolutely want to change something about it, it should be the `@return` docblock.

```suggestion
				return false;
```

src/bp-core/bp-core-rest-api.php
```diff
@@ -389,3 +361,64 @@ function bp_rest_register_field( $component_id, $attribute, $args = array(), $ob
...
+		return $result;
+	}
+
+	return new WP_Error(
+		'rest_no_route',
+		__( 'The V1 of the BuddyPress REST API is no longer supported, use the V2 instead.', 'buddypress' ),
+		array( 'status' => 404 )
+	);
+}
+add_filter( 'rest_pre_dispatch', 'bp_rest_api_v1_dispatch_error', 10, 3 );
```
imath, line 394: Excellent work here 💪 ! I confirm it behaves as expected.

imath (commented):
Hi @renatonascalves 

Thanks a lot for your work on this. First I'm sorry I'm hesitating a lot about the best way to deprecate v1. 

As I said in one of my review comments and considering the one you made on the Notices PR, I believe we should make sure to inform developers if:
- they are using directory v1 classes without having the BP REST plugin active => `_deprecated_class()` see the comment I made about `class-buddypress.php`
- they are including v1 files inside BuddyPress => `_deprecated_file()` see the comment I made about `bp-activity/classes/class-bp-rest-activity-endpoint.php`

This means renaming all v2 files and classes so that they match the autoloader regular map. Eg: `class BP_Activity_REST_Controller`=> class-bp-activity-rest-controller.php` (like you previously did I believe - sorry for changing my mind!!). As I've tested it, you'll find below the diff I added to my local test of your PR below (if it can save you some time).
[337.unfinished.patch](https://github.com/user-attachments/files/16856222/337.unfinished.patch)

The Manage Group Members UI using the REST API works as expected after the change. The `buddypress/v1` deprecated notice works as expected 👏 .

On a site note, once REST API v2 will be SVN committed, we'll need to update all BP Blocks to use the new rest version as `buddypress/v1` is hardcoded into these.

renatonascalves:
@imath Three clarifying questions:

Looking at this comment related to the `_deprecated_class` and it seems the location this should be added is inside a class constructor. Which is different from the the location `class-bp-patch` has it on.

https://github.com/WordPress/wordpress-develop/blob/9ad162e53ee9732c8ed63150650f7f9d3067a9d7/src/wp-includes/functions.php#L5662-L5663

I do [see some places](https://github.com/search?q=_deprecated_class%28+language%3APHP&ref=opensearch&type=code&l=PHP) where folks are using it outside of the constructor, but it doesn't seem to be the location it was _designed_ to be used.

---

Some of the new classes names are a bit off, like `BP_Members_REST_Notices_Controller` instead of `BP_Members_Notices_REST_Controller`. Is that expected?

---

> I would add to all these v1 named files a deprecation notice in case i**t's directly required by a plugin.**

I'm really curious if you know of a plugin that's actually loading one of our controller files manually into their own plugin.

That would be very odd, it's as if we would load the file for the `WP_REST_Controller` class so that we extend it, which would be uncommon to say the least. That technically would mean a conflict since BuddyPress and the other plugin would be loading the classes twice. So that doesn't make much sense to me. 🤔 

A [quick search](https://github.com/search?q=classes%2Fclass-bp-rest-components-endpoint.php+language%3APHP&ref=opensearch&type=code&l=PHP) came up empty but I assume you might know of an example.

And just to be clear, I'm not against this change. I'm mostly curious, would be flabbergasted, if someone would be doing that at all for this type of class/endpoint/controller.

src/bp-core/bp-core-rest-api.php
```diff
@@ -389,3 +361,64 @@ function bp_rest_register_field( $component_id, $attribute, $args = array(), $ob
...
+	wp_localize_script(
+		'bp-api-request',
+		'bpApiSettings',
+		array(
+			'unexpectedError'   => __( 'An unexpected error occurred. Please try again.', 'buddypress' ),
+			'deprecatedWarning' => __( 'The bp.apiRequest function is deprecated since BuddyPress 10.0.0, please use wp.apiRequest instead.', 'buddypress' ),
+		)
+	);
+}
+add_action( 'bp_init', 'bp_rest_api_register_request_script' );
```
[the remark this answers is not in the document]
renatonascalves: Right! I was going to apply this change in a second pr, separate ticket.

src/bp-core/bp-core-rest-api.php
```diff
@@ -389,3 +361,64 @@ function bp_rest_register_field( $component_id, $attribute, $args = array(), $ob
...
+
+	wp_localize_script(
+		'bp-api-request',
+		'bpApiSettings',
+		array(
+			'unexpectedError'   => __( 'An unexpected error occurred. Please try again.', 'buddypress' ),
+			'deprecatedWarning' => __( 'The bp.apiRequest function is deprecated since BuddyPress 10.0.0, please use wp.apiRequest instead.', 'buddypress' ),
+		)
+	);
+}
```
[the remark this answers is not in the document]
renatonascalves: I was going to apply this change in a second pr, separate ticket.

src/bp-core/bp-core-functions.php
renatonascalves: Ops! Will revert that.

src/bp-activity/classes/class-bp-rest-activity-endpoint.php
```diff
@@ -0,0 +1,1606 @@
+<?php
```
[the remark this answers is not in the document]
renatonascalves: If I understood correctly, then, we should not add this to the BP REST plugin files too, right?

src/bp-blogs/bp-blogs-template.php
renatonascalves: Created: https://buddypress.trac.wordpress.org/ticket/9228

src/class-buddypress.php
```diff
@@ -788,9 +779,9 @@ public function autoload( $class ) {
 		$class = strtolower( str_replace( '_', '-', $class ) );
 
 		if ( 'bp-rest-attachments' === $class ) {
-			$path = dirname( __FILE__ ) . "/bp-{$component}/classes/trait-attachments.php";
+			$path = __DIR__ . "/bp-{$component}/classes/trait-attachments.php";
 		} else {
-			$path = dirname( __FILE__ ) . "/bp-{$component}/classes/class-{$class}.php";
+			$path = __DIR__ . "/bp-{$component}/classes/class-{$class}.php";
 		}
```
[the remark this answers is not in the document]
renatonascalves: I like this. I only have one question regarding _how_ we are adding the `_deprecated_class`.

src/bp-activity/classes/class-bp-rest-activity-endpoint.php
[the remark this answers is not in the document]
imath: Yes, it's only in the case a plugin developer would directly require this file (I agree there's few chances one would... but we never know).

imath:
Hi @renatonascalves about the 3 clarifying questions:

1. If `_deprecated_class` is not the right function to use, I'm fine with a `_doing_it_wrong()` informing to use v2 controllers instead.
2. About class names, the most important is we make sure to have `BP_{component_id}...` to be inside the regular map of the autoloader. Your way of naming it with `REST_Controller` at the end is better I agree.
3. If you thing the above point is enough, I trust you. My reasoning is more "file was there in 14.0, it's no more the case in 15.0, use a deprecated function to inform about it just in case" 😅

renatonascalves:
@imath

1. `_deprecated_class` will be moved to the BP REST plugin classes constructor.
2. I'll update with the suggested naming so that `REST_Controller` is used at the end.
3. That sounds good. We can keep it.

imath:
Awesome, maybe we still need to inform about deprecated class usage from BuddyPress when the BP REST plugin is not active?
At least I believe we should remove these class names from the autoloader irregular map.

renatonascalves:
@imath

- Removed V1 class names from the autoloader irregular map (the BP REST plugin are loading them now);
- V2 classes uses the autoloader, as extected;
- Pr updated with the new class naming `_REST_Controller` at the end;
- Added a deprecation notice for v1 named files in case it's directly required by a plugin;
- `_deprecated_class` moved to the class constructor of the BP REST plugin endpoints . The error appears in the `debug.log` and in the header of the HTTP response.

> maybe we still need to inform about deprecated class usage from BuddyPress when the BP REST plugin is not active?

This is not necessary because without the BP REST plugin, a site can't use the V1 endpoints at all. They can only use the V2 endpoints. And if the need to use V1, they need to install the BP REST plugin.

Here is the PR in the BP REST plugin: https://github.com/buddypress/BP-REST/pull/512

:)

src/bp-core/classes/trait-attachments.php
```diff
@@ -139,6 +139,7 @@ protected function upload_cover_from_file( $file ) {
 		}
 
 		return sprintf(
+			/* translators: %1$s: baseurl, %2$s: subdir, %3$s: basename. */
```
imath, line 142:
```suggestion
```
Nothing is translated there, we don't need this comment imho.

imath (approved):
Hi @renatonascalves congrats for the achieved works! Amazing job.

I've added a minor suggestion about translators inline comment. Maybe we should rename PHPUnit REST testcases / files to follow the `controller` switch, eg: ` BP_Test_REST_Group_Endpoint` => `BP_Test_Group_REST_Controller` & `test-endpoint.php` => `test-controller.php`?

Feel free to SVN Commit!
NB: As you're adding a lot of files, don't forget to `svn stat` and eventually `svn add` new files before committing.

💚 💪

renatonascalves:
@imath Awesome! I'll apply the latest suggestions. I actually forgot to rename the phpunit test files related to the endpoints to controller. Good catch!

src/bp-core/classes/trait-attachments.php
renatonascalves: Good catch!

src/bp-activity/classes/class-bp-rest-activity-endpoint.php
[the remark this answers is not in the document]
renatonascalves: Added! ✅

src/class-buddypress.php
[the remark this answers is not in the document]
renatonascalves: ✅

PR #341, opened 2024-07-20
Block user from constantly request activation email resend

renatonascalves:
Here I'm suggesting an approach to block users from constantly request the resending of activation emails. This approach is filterable and flexible. And accounts for both the web version and the REST API.

Trac ticket: https://buddypress.trac.wordpress.org/ticket/9137

---

### DIFF
--- a/src/bp-members/classes/class-bp-signup.php
+++ b/src/bp-members/classes/class-bp-signup.php
@@ -3,7 +3,7 @@
  * Signups Management class.
  *
  * @package BuddyPress
- * @subpackage coreClasses
+ * @subpackage Signup
  * @since 2.0.0
  */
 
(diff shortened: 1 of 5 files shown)


src/bp-members/classes/class-bp-signup.php
```diff
@@ -1084,4 +980,161 @@ public static function delete( $signup_ids = array() ) {
...
+
+		/**
+		 * Filters whether an activation email can be resent.
+		 *
+		 * @since 15.0.0
+		 *
+		 * @param bool      $retval Whether an activation email can be resent.
+		 * @param BP_Signup $signup The signup object.
+		 */
+		return apply_filters( 'bp_core_signup_resend_activation', $retval, $signup );
```
renatonascalves, line 1138: I chose not to name this `bp_core_signup_resend_activation_email`, because a site could be using a different type of notification.

src/bp-members/classes/class-bp-signup.php
```diff
@@ -1084,4 +980,161 @@ public static function delete( $signup_ids = array() ) {
...
+
+	/**
+	 * Check if an activation email can be resent.
+	 *
+	 * @since 15.0.0
+	 *
+	 * @param BP_Signup $signup The signup object.
+	 * @return bool
+	 */
+	public static function resend_activation( $signup ) {
```
dcavins, line 1102: The function name makes this sound like an action and not a check. Maybe `allow_activation_resend()` would be clearer (and less close to `resend()`)?

renatonascalves: Agree! Will update.

PR #350, opened 2024-07-25
Add debounce to @mentions remote filter for improved performance

Praveenkroot:
Add debounce function to improve performance of @mentions remote filter

- Integrated debounce function to throttle the frequency of remoteFilter calls.
- Enhanced user experience by reducing the number of AJAX requests made during typing.
- Updated the remoteFilter callback in mentions.js to use debounce.
- Adjusted debounce delay to 300ms to balance responsiveness and performance.

This change aims to optimize the performance of the @mentions feature by minimizing unnecessary AJAX requests and improving overall responsiveness.

Trac ticket: 

---

### DIFF
--- a/src/bp-activity/js/mentions.js
+++ b/src/bp-activity/js/mentions.js
@@ -18,22 +18,42 @@ window.bp = window.bp || {};
 	 * of AJAX requests for user mentions.
 	 *
 	 * This function ensures that the `func` is only called after `wait` milliseconds
-	 * have passed since the last time it was invoked. This helps reduce the frequency
-	 * of AJAX requests and improves the efficiency of the @mentions feature.
+	 * have passed since the last invocation, with an optional immediate execution.
+	 * This helps reduce the frequency of AJAX requests and improves the efficiency of the @mentions feature.
 	 *
 	 * @param {Function} func The function to debounce.
 	 * @param {number} wait The number of milliseconds to wait before calling `func`.
+	 * @param {boolean} [immediate=true] Whether to call the function immediately.
 	 * @returns {Function} A debounced version of the provided function.
 	 */
-	function debounce(func, wait) {
-	    let timeout;
-	    return function() {
-	        const context = this;
-	        const args = arguments;
-	        clearTimeout(timeout);
-	        timeout = setTimeout(() => func.apply(context, args), wait);
-	    };
-	}
+	function debounce(func, wait, immediate) { 
+		'use strict'; 
+	 
+		var timeout; 
+		wait      = (typeof wait !== 'undefined') ? wait : 20; 
+	        immediate = (typeof immediate !== 'undefined') ? immediate : true; 
+	 
+		return function() { 
+	 
+			var context = this, args = arguments; 
+			var later = function() { 
+				timeout = null; 
+	 
+				if (!immediate) { 
+					func.apply(context, args); 
+				} 
+			}; 
+	 
+			var callNow = immediate && !timeout; 
+	 
+			clearTimeout(timeout); 
+			timeout = setTimeout(later, wait); 
+	 
+			if (callNow) { 
+				func.apply(context, args); 
+			} 
+		}; 
+	} 
 
 
 	/**
@@ -188,6 +208,12 @@ window.bp = window.bp || {};
 				remoteFilter: debounce(function( query, render_view ) {
 					var self = $( this ),
 						params = {};
+					
+					// Skip AJAX request if query is empty
+					if (!query.trim()) {
+				        	render_view([]); // Return an empty array to indicate no results
+				        	return;
+				    	}
 
 					mentionsItem = mentionsQueryCache[ query ];
 					if ( typeof mentionsItem === 'object' ) {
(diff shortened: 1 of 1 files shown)


src/bp-activity/js/mentions.js
```diff
@@ -13,6 +13,29 @@ window.bp = window.bp || {};
...
+	 *
+	 * This function ensures that the `func` is only called after `wait` milliseconds
+	 * have passed since the last time it was invoked. This helps reduce the frequency
+	 * of AJAX requests and improves the efficiency of the @mentions feature.
+	 *
+	 * @param {Function} func The function to debounce.
+	 * @param {number} wait The number of milliseconds to wait before calling `func`.
+	 * @returns {Function} A debounced version of the provided function.
+	 */
+	function debounce(func, wait) {
```
renatonascalves, line 28: We actually have some prior art. Could we add this in a way that can be reshared?

https://github.com/buddypress/buddypress/blob/151f0938e05ac357864cb6e6b87535f935db0808/src/bp-templates/bp-nouveau/js/buddypress-priority-menu.js#L17-L44

renatonascalves (changes requested):

renatonascalves:
A trac ticket would be great too.

src/bp-activity/js/mentions.js
Praveenkroot: Hi @renatonascalves 

I've implemented the changes:
1. Updated the debounce function to ensure consistency with existing code.
2. Added functionality to skip AJAX requests if the query is empty.

Please review the latest commit and let me know if there are any further adjustments needed.

renatonascalves: I actually meant that we could reuse the same code in both places.

Praveenkroot: Thank you for the clarification. I understand your suggestion to reuse the same debounce function across different parts of the project. To achieve this, I propose extracting the debounce function from **buddypress/src/bp-templates/bp-nouveau/js/buddypress-priority-menu.js** and placing it in a shared location that can be imported where needed.

Specifically, we can create a common utility file, **utils.js**, where we define the debounce function. Then, we can update both **mentions.js** and **buddypress-priority-menu.js** to import and use the debounce function from **utils.js**. This approach ensures consistency and reduces redundancy, as any updates to the debounce function will only need to be made in one place.

Would you agree with this approach? If so, could you please advise on the best location to place this shared function? I'll go ahead and implement these changes once I have your approval.

renatonascalves: @imath Do you have suggestions?

src/bp-activity/js/mentions.js
```diff
@@ -13,6 +13,49 @@ window.bp = window.bp || {};
...
+	 * @param {number} wait The number of milliseconds to wait before calling `func`.
+	 * @param {boolean} [immediate=true] Whether to call the function immediately.
+	 * @returns {Function} A debounced version of the provided function.
+	 */
+	function debounce(func, wait, immediate) { 
+		'use strict'; 
+	 
+		var timeout; 
+		wait      = (typeof wait !== 'undefined') ? wait : 20; 
+	        immediate = (typeof immediate !== 'undefined') ? immediate : true;
```
renatonascalves, line 34: Looks like there is some spacing issues here.

src/bp-activity/js/mentions.js
```diff
@@ -162,9 +205,15 @@ window.bp = window.bp || {};
...
 				 */
-				remoteFilter: function( query, render_view ) {
+				remoteFilter: debounce(function( query, render_view ) {
 					var self = $( this ),
 						params = {};
+					
+					// Skip AJAX request if query is empty
+					if (!query.trim()) {
+				        	render_view([]); // Return an empty array to indicate no results
+				        	return;
```
renatonascalves, line 215: Here too.

renatonascalves (changes requested):

renatonascalves:
@Praveenkroot Let's not try to reuse this functionality. It might be harder than I expected. And might require a large conversation. We can fix the issue here with the current implementation.

src/bp-activity/js/mentions.js
Praveenkroot: I've updated the spacing issues.

Praveenkroot:
@renatonascalves Thank you for the feedback. I understand that reusing the functionality might be complex and could require extensive discussion. I’ll focus on addressing the issue with the current implementation as you suggested.

I will create a ticket on BuddyPress Trac regarding this issue and provide a link to it here.

Thanks for your guidance!

Praveenkroot:
> @renatonascalves Thank you for the feedback. I understand that reusing the functionality might be complex and could require extensive discussion. I’ll focus on addressing the issue with the current implementation as you suggested.
> 
> I will create a ticket on BuddyPress Trac regarding this issue and provide a link to it here.
> 
> Thanks for your guidance!

renatonascalves:
@Praveenkroot I think this pr is only pending a ticket with more information and possibly some testing steps.

Praveenkroot:
Thank you for your patience. I've created the Trac ticket as requested, which includes additional details and testing steps for the issue we discussed.

You can find the ticket here: https://buddypress.trac.wordpress.org/ticket/9239#ticket

Please let me know if there’s anything else you'd like to see added or adjusted.

Thanks again for your guidance and feedback!

renatonascalves:
Thanks @Praveenkroot 🤟🏽

PR #352, opened 2024-07-26
Stop including 12.0 deprecated functions by default when initial BP version is 15.0 or up

imath:
Always Include 12.0 deprecated functions as soon as initial version is < 15.0.0.

Trac ticket: https://buddypress.trac.wordpress.org/ticket/9210

---


tests/phpunit/testcases/core/functions.php
```diff
@@ -878,59 +878,85 @@ public function override_initial_version() {
...
-
-		add_filter( 'pre_option__bp_initial_major_version', array( $this, 'override_initial_version' ), 10, 0 );
+		// First BuddyPress installed version is older than 11.0. In this case, the initial version is 0.
+		// See https://buddypress.trac.wordpress.org/changeset/13304
+		$this->bp_initial_version = 0;
 
 		$versions = bp_get_deprecated_functions_versions();
 
-		remove_filter( 'pre_option__bp_initial_major_version', array( $this, 'override_initial_version' ), 10, 0 );
+		// Include 12.0 until version 16.0.
```
renatonascalves, line 928: I think you meant 15 here?
```suggestion
		// Include 12.0 until version 15.0.
```

tests/phpunit/testcases/core/functions.php
```diff
@@ -878,59 +878,85 @@ public function override_initial_version() {
...
+
+		// Testing with previous major version.
+		$this->bp_initial_version = $current_version - 1;
 
-		$this->assertTrue( 2 === count( $versions ) );
+		$versions = bp_get_deprecated_functions_versions();
 
-		// Even if this version does not exist in deprecated functions files, we should load the 2 lasts.
-		$this->bp_initial_version = '1.0';
+		// Include 12.0 until version 16.0.
```
renatonascalves, line 940: I think you meant 15 here?
```suggestion
		// Include 12.0 until version 15.0.
```

tests/phpunit/testcases/core/functions.php
```diff
@@ -878,59 +878,85 @@ public function override_initial_version() {
...
+		}
 
-		add_filter( 'pre_option__bp_initial_major_version', array( $this, 'override_initial_version' ), 10, 0 );
+		// Testing with 2 versions older.
+		$this->bp_initial_version = (float) 15 === $current_version ? (float) 12 : $current_version - 2;
 
 		$versions = bp_get_deprecated_functions_versions();
 
-		remove_filter( 'pre_option__bp_initial_major_version', array( $this, 'override_initial_version' ), 10, 0 );
+		// Include 12.0 until version 16.0.
```
renatonascalves, line 952: I think you meant 15 here?

```suggestion
		// Include 12.0 until version 15.0.
```

tests/phpunit/testcases/core/functions.php
```diff
@@ -878,59 +878,85 @@ public function override_initial_version() {
...
 
-		$versions = bp_get_deprecated_functions_versions();
-
-		$this->assertContains( 12.0, $versions, '12.0 deprecated functions should be loaded when 12.1.1 was the first installed version.' );
+		// Fresh install.
+		$this->bp_initial_version = $current_version;
+		$expected_versions        = array();
 
-		$this->bp_initial_version = '9.0';
+		// Include 12.0 until version 16.0.
```
renatonascalves, line 910: I think you meant 15 here?
```suggestion
		// Include 12.0 until version 15.0.
```

tests/phpunit/testcases/core/functions.php
```diff
@@ -878,59 +878,85 @@ public function override_initial_version() {
...
+	 * This test acts like a reminder: $deprecated_functions_versions needs to include each versions of the `/src/bp-core/deprecated` directory.
+	 *
+	 * eg:  if `/src/bp-core/deprecated/15.0.php` exists, then $deprecated_functions_versions should include `15.0`.
+	 *
+	 * @group bp_get_deprecated_functions_versions
+	 * @ticket BP9210
+	 */
+	public function test_bp_get_deprecated_functions_versions_array_is_consistent_with_deprecated_files() {
+		// When current major version is the initial version, we shouldn't load any deprecated functions file.
+		$this->assertTrue( false !== bp_get_deprecated_functions_versions(), 'Please check the list of `$deprecated_functions_versions` in `bp_get_deprecated_functions_versions()`. There should be one for each file of the `/src/bp-core/deprecated` directory.' );
```
renatonascalves, line 890: I think now would be a good time to add more test coverage. Like testing this behavior, https://github.com/buddypress/buddypress/blob/8cfca1802adeeb85fff69b8848deec6996d49e39/src/bp-core/bp-core-functions.php#L4975-L4977, and others.

renatonascalves: I can add those tests separately too. If necessary.

src/bp-core/bp-core-functions.php
```diff
@@ -5012,27 +5012,20 @@ function( $file ) {
...
-	}
-
-	$index_first_major = array_search( $initial_version, $deprecated_functions_versions, true );
-	if ( false === $index_first_major ) {
-		return array_splice( $deprecated_functions_versions, -2 );
-	}
-
-	$latest_deprecated_functions_versions = array_splice( $deprecated_functions_versions, $index_first_major );
+		// Load 12.0 deprecated functions only when BP was installed with 12.0, 14.0 or 15.0.
+		if ( in_array( $initial_version, array( 12.0, 14.0, 15.0 ) ) ) {
```
renatonascalves, line 5016:
```suggestion
		if ( in_array( $initial_version, array( 12.0, 14.0, 15.0 ), true ) ) {
```

renatonascalves (changes requested):

emaralive:
I see that the filter hook from the proposed-new-9210.01.patch was dropped with no explanation as to why. The way I see it, as it stands, with a development version the only choices for loading deprecated functions is to:

1. Don't load any deprecated functions.
2. Load all deprecated functions.

Which means reliance on the unit tests to catch issues, and in this case, seem to have been unreliable. Which is why I added the hook to allow for a 3rd option of allowing deprecated functions for development version to load the same as a production version which represented a manual fail safe check. Perhaps, someone could explain.

imath:
Sorry @emaralive I thought this part of my comment on the Trac ticket was explaining the scandir that is only done during unit tests:
> -
`test_bp_get_deprecated_functions_versions_array_is_consistent_with_deprecated_files()`
is there to check that when we create a new `/src/bp-
core/deprecated/X.X.php` file we don't forget to update the
`$deprecated_functions_versions` to include `X.X` version.

Deprecated code is not not loaded into this function but here: https://github.com/buddypress/buddypress/blob/8cfca1802adeeb85fff69b8848deec6996d49e39/src/class-buddypress.php#L652

imath
https://imathi.eu
________________________________
De : emaralive ***@***.***>
Envoyé : Tuesday, August 6, 2024 3:31:52 PM
À : buddypress/buddypress ***@***.***>
Cc : imath ***@***.***>; Author ***@***.***>
Objet : Re: [buddypress/buddypress] Stop including 12.0 deprecated functions by default when initial BP version is 15.0 or up (PR #352)


I see that the filter hook from the proposed-new-9210.01.patch was dropped with no explanation as to why. The way I see it, as it stands, with a development version the only choices for loading deprecated functions is to:

  1.  Don't load any deprecated functions.
  2.  Load all deprecated functions.

Which means reliance on the unit tests to catch issues, and in this case, seem to have been unreliable. Which is why I added the hook to allow for a 3rd option of allowing deprecated functions for development version to load the same as a production version which represented a manual fail safe check. Perhaps, someone could explain.

tests/phpunit/testcases/core/functions.php
imath: That would be great 💪

renatonascalves (approved):

imath:
> I see that the filter hook from the proposed-new-9210.01.patch was dropped with no explanation as to why. The way I see it, as it stands, with a development version the only choices for loading deprecated functions is to:
> 
> 1. Don't load any deprecated functions.
> 2. Load all deprecated functions.
> 
> Which means reliance on the unit tests to catch issues, and in this case, seem to have been unreliable. Which is why I added the hook to allow for a 3rd option of allowing deprecated functions for development version to load the same as a production version which represented a manual fail safe check. Perhaps, someone could explain.

Hi @emaralive sorry it took me a while to understand your point, here's the explanation and I'm open to add a filter if needed:
https://buddypress.trac.wordpress.org/ticket/9224

See #358

PR #362, opened 2024-08-10
Raise WordPress required version to 6.4

renatonascalves:
Trac ticket: https://buddypress.trac.wordpress.org/ticket/9225

---


bp-loader.php
```diff
@@ -19,20 +19,20 @@
  * License URI:       https://www.gnu.org/licenses/gpl-2.0.html
  * Text Domain:       buddypress
  * Domain Path:       /bp-languages/
- * Requires PHP:      5.6
- * Requires at least: 6.1
+ * Requires PHP:      7.0
```
renatonascalves, line 22: According to [WordPress stats](https://wordpress.org/about/stats/), PHP 7.0 has less than 2% of usage. PHP 7.3 has around ~3.8%.

And PHP 7.4, 37.6%. **I personally would like to drop support for <7.3**. Meaning that BuddyPress would require 7.4>.

It's not clear to me why we were supporting 5.6, but we don't have data specific to BuddyPress, as far as I know.

renatonascalves: 7.0 is the minumum version supported by 6.4, that's why I defaulted to 7.0.

.github/workflows/unit-test-object.yml
```diff
@@ -11,14 +11,34 @@ on:
...
-        object: ["false", "memcached", "redis"]
+        include:
+          - php: '8.3'
+            wp_version: 'trunk'
+            object: 'redis'
+            multisite: true
+          - php: '8.3'
+            wp_version: 'trunk'
+            object: 'memcached'
+            multisite: false
```
renatonascalves, line 25: Those two are essentially to test with the object cache services.

.github/workflows/unit-test-object.yml
```diff
@@ -11,14 +11,34 @@ on:
...
+            object: ''
+            multisite: false
+          - php: '8.2'
+            wp_version: 'latest'
+            object: ''
+            multisite: false
+          - php: '8.1'
+            wp_version: 'latest'
+            object: ''
+            multisite: false
```
renatonascalves, line 37: PHP 8.1 and 8.2 are used the most by WordPress installations: https://wordpress.org/about/stats/

renatonascalves:
@imath Could you take a look at the matrix here? I tried to balance our need with the Github Action limits. We want to run as many scenarios as possible, but at the same time, we don't have unlimited resources.

.github/workflows/unit-test-object.yml
```diff
@@ -11,14 +11,34 @@ on:
...
+          - php: '7.4'
+            wp_version: 'latest'
+            object: ''
+            multisite: false
```
renatonascalves, line 41: 7.4 accounts for 37% of all WordPress installations. Maybe we should test this more?!

bp-loader.php
imath: I believe we should follow WordPress requirements about this. Raising PHP version according to the minimum required WP version: WP 6.4 => PHP 7.0 is fine with me.
WP 6.6 is supporting back to [7.2](https://core.trac.wordpress.org/browser/branches/6.6/src/wp-includes/version.php)

src/readme.txt
```diff
@@ -4,10 +4,10 @@ Donate link:       https://wordpressfoundation.org
 Tags:              community, group, members, network, self-hosted
 License:           GNU General Public License v2 or later
 License URI:       https://www.gnu.org/licenses/gpl-2.0.html
-Requires PHP:      5.6
-Requires at least: 6.1
+Requires PHP:      7.0
+Requires at least: 6.4
 Tested up to:      6.6
-Stable tag:        14.0.0
+Stable tag:        15.0.0-alpha
```
imath, line 10:
```suggestion
Stable tag:        14.0.0
```
Please do not change stable tag. This only happens when building a new minor/major version. See https://github.com/buddypress/buddypress/blob/master/docs/contributor/project/release/build-checklist.md#version-bumps-in-dev

src/readme.txt
```diff
@@ -54,15 +54,15 @@ Growing the BuddyPress community means better software for everyone!
 
 To run BuddyPress, we recommend your host supports:
 
-* PHP version 7.4 or greater.
+* PHP version 8.1 or greater.
```
imath, line 57:
```suggestion
* PHP version 7.4 or greater.
```

Let's follow how WordPress evolves here too: [7.4 is still minimum recommended PHP version](https://wordpress.org/about/requirements/). Let's keep it this way.

src/readme.txt
```diff
@@ -54,15 +54,15 @@ Growing the BuddyPress community means better software for everyone!
...
 * MySQL version 5.7 or greater OR MariaDB version 10.4 or greater.
```
imath, line 58:
```suggestion
* MySQL version 8.0 or greater OR MariaDB version 10.5 or greater.
```

I think we should also follow WordPress db recommended versions.

.github/workflows/unit-test-object.yml
imath, line 41:
```suggestion
          - php: '7.4'
            wp_version: '6.4'
            object: ''
            multisite: false
         - php: '7.4'
            wp_version: '6.4'
            object: ''
            multisite: true
```
I think we should test PHP 7.4 with WP 6.4 + add a job to test multisite.

imath (changes requested):
Thanks a lot for your work on it @renatonascalves I've added some comments in the code. I think we should follow WordPress recommended versions and choose PHP required version according to the first WP version we support.

src/readme.txt
renatonascalves: Ops! Good catch!

src/readme.txt
renatonascalves: This is actually the recommended PHP version, not the minumum.

renatonascalves: Ops. I see. That's correct!

.github/workflows/unit-test-object.yml
```diff
@@ -11,14 +11,38 @@ on:
...
+          - php: '7.4'
+            wp_version: '6.4'
+            object: ''
+            multisite: false
+         - php: '7.4'
+            wp_version: '6.4'
+            object: ''
+            multisite: true
```
renatonascalves, line 45:
```suggestion
          - php: '7.4'
            wp_version: '6.4'
            object: ''
            multisite: false
          - php: '7.4'
            wp_version: '6.4'
            object: ''
            multisite: true
```

imath (approved):
Thanks a lot! All lights green to me, feel free to SVN commit it 💪

renatonascalves:
Thanks! I'll publish a post in the bpdevel blog too about it. 🎈

PR #368, opened 2024-09-01
Revamped Notices feature

imath:
The goal of this PR is to transform the code used into the Private Messages component (*an optional component => only available when the Private Messages component is active*) to send all community members a “SiteWide Notice” into a Members component (*a required component => always available*) feature so that:

- The BP Team can use it to inform Administrators about important changes: this new Members Feature is replacing the Admin Notifications workaround we introduced in 11.4.0 to communicate about 12.0.0 important changes.
- Site administrators can notice all community members or all contributors or all other administrators about important informations dealing with community, contribution, or administration matters.
- Integration with the Notifications component is improved as Notices behave similarly to Notifications => they “disappear” once they are dismissed by the user.
## Notices: a Members component feature

While I first thought Notices should be a Core component feature, I quickly changed my mind when working on code changes. The Core component is not setting [needed globals](https://github.com/buddypress/buddypress/blob/master/src/bp-core/classes/class-bp-component.php#L345|L448) such as DB table/table meta and after more thoughts I believe Notices have more to do with Members.

The Members Notices can be deactivated using this filter:

```php
add_filter( 'bp_is_members_notices_active', '__return_false' );
```

Although I’d strongly recommend not to disable it, we’ll always make some adventure users unhappy with it…

A notice is a piece of informations targeting specific member roles:

- `community` All members are targeted by the notice.
- `contributors` Only members able to write posts are targeted by the notice.
- `admins` Only administrators are targeted by the notic.

A notice can have 5 priority levels. 2 levels are not available to administrators:

- `0`  the “Very High” priority level is restricted to the BuddyPress team to inform Administrators about important plugin changes.
- `1` the “High” priority level can be used to move a notice to the top of the notices list.
- `2` the “Regular” priority level is the default one.
- `3` the “Low” priority level can be used to move a notice at the bottom of the notices list.
- `127` is a specific level used to deactivate a notice (just like it was possible within the Private Messages component)

A notice can include a call to action as long as the destination URL has the same domain name than the WordPress site.

Notices are stored into the `wp_bp_notices` table and data about the list of users who dismissed a notice is no longer inside a serialized array of the corresponding user meta but stored as rows into the `wp_bp_notices_meta` table. There’s one row for a user/notice and the name of the Notice meta is `dismissed_by`: Notice ID (`notice_id`  field of the table) was dismissed by (`meta_key` field of the table) User ID (`meta_value`  field of the table).
The BuddyPress DB Schema was updated to migrate data inside the `wp_bp_messages_notices` table if it exists into the new `wp_bp_notices` table.

**Notices management**
Administrators can manage notices from the `Users > Member notices` WP Dashboard menu. From there they can add, edit, deactivate or delete all the notices they created. BuddyPress “Very High” notices cannot be edited or created by the administrators. They can deactivate this kind of notices but they need to wait for all administrators to dismiss it before being able to delete it.

**The BP Notices center**
An orange bell with the number of unread notices/notifications is added to the WP Admin Bar only if there is a notice or a notification (when the Notifications component is active) to display.

Once the orange bell is clicked an HTML Popover is appearing: the first part of it is displaying a “Community notices” section and when the Notifications component is active, it also includes a “Personal notifications” section.

> [!NOTE]
> If users browser doesn’t support the [HTML Popover API](https://caniuse.com/?search=popover), they are redirected to the Notices Member’s front-end profile area when clicking on the orange bell.

> [!NOTE]
> In other words, the Notifications component is using this BP Notices Center to display the logged in member’s notifications. It doesn’t create any new node to the WP Admin Bar anymore and the `src/bp-notifications/bp-notifications-adminbar.php` file is deprecated since 15.0.0.

The “Community notices” section is a “no JS” slider allowing the logged in member to slide up to 5 top priority notices. If there are more than 5 active notices a link is added to the slider’s pagination to view all notices from the Notices Member’s front-end profile area.

A notice can always be dismissed from there thanks to the corresponding “Dismiss” button. If the notice contains a call to action, another button is added for it and clicking on this button will first dismiss the notice before redirecting the user to the call to action URL.

BuddyPress is using this call to action when noticing an upgrade to a major release was successful so that the administrator can directly open the BP Hello/changelog modal to discover what’s new. As a reminder this modal is only automatically displayed after a successful install.

> [!NOTE]  
> BuddyPress Admin Notices are still displayed into the Notices tab of the BuddyPress settings administration screen. From there it’s possible to view dismissed notices as well.

**Notices Member’s front-end profile area**
All notices are by default available into a Notices navigation of the user’s profile. When the Notifications component is active this navigation is moved as a Notifications sub navigation. In this case the Notifications count also takes in account the Notices count. The Notices loop is paginated 5 by 5 and can be filtered to list all unread or dismissed notices. Both Legacy and Nouveau template packs were updated to include notices templates.

## The component’s feature API (new!)

The PR includes a new `BP_Component_Feature` class. Looking back at how we implemented the Site Invitations feature, I believed this part needed to be improved. For Site Invitations we created a component using the `BP_Members_Invitations_Component` object just to be able to have a primary nav into the member’s profile area. The downside of this is it’s creating an active component into the `buddypress()` global. Extending this new `BP_Component_Feature` class which itself extends the `BP_Component` class we avoid the downside and have better control on how to build feature specificities (see `BP_Members_Notices_Feature`). I believe we should go a bit further grouping features files into a feature’s directory. For now I just did the same than we did before: group all functions, template tags etc.., into a single `bp-members-notices.php` file (`bp-{component-name}-{feature-name}.php`).

## The Notices Rest Controller

> [!TIP]
> Until BP REST API v2 is merged into Core, you’ll need to use the following code snippet within a [bp-custom.php](https://github.com/buddypress/buddypress/blob/master/docs/user/advanced/functionalities.md) file to be able to test the JavaScript Powered dismiss action:

```php
function force_notices_feature_rest_api( $feature_class ) {
    add_action( 'bp_rest_api_init', array( $feature_class, 'rest_api_init' ), 10 );
}
add_action( 'bp_members_notices_setup_actions', 'force_notices_feature_rest_api' );
```

I remember having discussions with @renatonascalves about the best way to deprecate BP REST API v1 and class naming for the BP REST API v2. I was confronted to it and after more thoughts, I believe we should name v2 REST API Class controllers like this `BP_{component}_{feature}_REST_Controller` (I probably changed my mind there, sorry!): one of benefit for this is we are following the class autoloader regular map and don’t need to add entries for irregular ones. Here’s the deprecation strategy I used:

- In `BuddyPress->autoload()` keep deprecated endpoint class, eg: `'BP_REST_Sitewide_Notices_Endpoint' => 'messages', // deprecated.`
- Keep & deprecate the old class file, see `src/bp-messages/classes/class-bp-rest-sitewide-notices-endpoint.php` for an example.

The `BP_Members_Notices_REST_Controller`  will greatly benefit from @renatonascalves review & improvements.

Thanks in advance for your tests & reviews!

> [!IMPORTANT]
> Go to `site.url/wp-admin/` first to test the PR, the upgrade routine needs to be accomplished first, you'd get DB errors otherwise.

Trac ticket: https://buddypress.trac.wordpress.org/ticket/9098

---


src/bp-members/classes/class-bp-members-notice.php
```diff
@@ -0,0 +1,758 @@
...
+ * BuddyPress Members Notice Class.
+ *
+ * @package buddypress\bp-members\classes\class-bp-members-notice
+ * @since 15.0.0
+ */
+
+// Exit if accessed directly.
+if ( ! defined( 'ABSPATH' ) ) {
+	exit;
+}
```
renatonascalves, line 12:
```suggestion
defined( 'ABSPATH' ) || exit;
```

src/bp-members/classes/class-bp-members-notice.php
```diff
@@ -0,0 +1,758 @@
...
+	 *     @type array        $exclude          An array of notice IDs to exclude from results. Defaults to `[]`.
+	 *     @type array        $target__in       An array of targeted audiences to include in results. Possible targets
+	 *                                          are 'community', 'contributors', 'admins'. Defaults to `[]`.
+	 *     @type null|integer $priority         Use `1` to get high priority notices, `2` to get regular ones or `3` to
+	 *                                          get low ones. `0` is restriced to BuddyPress usage: please don't use it.
+	 *                                          Defaults to `null`.
+	 *     @type boolean      $count_total_only `true` to only get the total number of notices, `false` otherwise.
+	 *                                          Defaults to `false`.
+	 * }
+	 * @return array|integer|null List of notices, total number of notices for a count only query or null if nothing was found.
```
renatonascalves, line 405: Most of our `::get`, returns the objects (or ids only) and the total number of matches objects at the same time. I think maybe to avoid a second query. Not all classes support this though.

src/bp-members/classes/class-bp-members-notice.php
```diff
@@ -0,0 +1,758 @@
...
+				$result = $wpdb->get_results(
+					"SELECT n.* FROM {$bp->members->table_name_notices} n
+					{$join_sql}
+					{$where_sql}
+					ORDER BY priority ASC, date_sent DESC
+					{$limit_sql}"
+				);
+
+				if ( $is_user_notices ) {
+					wp_cache_set( $args['user_id'], $result, 'bp_member_top_priority_notices' );
```
renatonascalves, line 534: As part of https://buddypress.trac.wordpress.org/ticket/8552, we should add support for `cache_results`. I can do it after this is merged too.

src/bp-members/classes/class-bp-members-notice.php
```diff
@@ -0,0 +1,758 @@
...
+	/**
+	 * Pulls up a list of notices.
+	 *
+	 * To get all notices, pass a value of -1 to pag_num.
+	 *
+	 * @global wpdb $wpdb WordPress database object.
+	 *
+	 * @since 15.0.0
+	 *
+	 * @param array $args {
```
renatonascalves, line 385: Is it possible to `orderby` by something? Priority maybe? Target?

src/bp-members/classes/class-bp-members-notice.php
```diff
@@ -0,0 +1,758 @@
...
+	 *     An array of arguments.
+	 *
+	 *     @type integer      $user_id          The user ID to get notices for. Defaults to `0`.
+	 *     @type integer      $pag_num          Number of notices per page. Defaults to `20`.
+	 *     @type integer      $pag_page         The page number to retrieve.  Defaults to `1`.
```
renatonascalves, line 390: Can get change this to a more familiar syntax? Like `$per_page` and `$page`.

src/bp-members/classes/class-bp-members-notice.php
renatonascalves: Also sort.

src/bp-members/classes/class-bp-members-notices-admin.php
```diff
@@ -0,0 +1,405 @@
+<?php
+/**
+ * BuddyPress members component Site-wide Notices admin screen.
+ *
+ * @package buddypress\bp-members\classes\class-bp-members-notices-admin
+ * @since 15.0.0
```
renatonascalves, line 6:
```suggestion
 * @package BuddyPress
 * @subpackage Members
 * @since 15.0.0
```

src/bp-members/classes/class-bp-members-notice.php
```diff
@@ -0,0 +1,758 @@
...
+}
+
+/**
+ * BuddyPress Members Notice Class.
+ *
+ * Use this class to create, activate, deactivate or delete notices.
+ *
+ * @since 15.0.0
+ */
+class BP_Members_Notice {
```
renatonascalves, line 21: I think this classs needs `#[AllowDynamicProperties]` too.

src/bp-members/classes/class-bp-members-notices-admin.php
```diff
@@ -0,0 +1,405 @@
...
+				array(
+					'id'       => 0,
+					'title'    => '',
+					'content'  => '',
+					'target'   => '',
+					'priority' => 2,
+					'url'     => '',
+					'text'     => '',
+				)
+			);
```
renatonascalves, line 142: While testing this, and looking at the code, it seems title and content should be required. But there is no _visual_ indication of that. Try to send a notice without any content and you get a successful message.

CleanShot 2024-09-02 at 13 14 08@2x

src/bp-members/classes/class-bp-members-notices-admin.php
```diff
@@ -0,0 +1,405 @@
...
+
+			<div id="col-container" class="wp-clearfix">
+				<div id="col-left">
+					<div class="col-wrap">
+						<?php $this->notice_form(); ?>
+					</div>
+				</div><!-- /col-left -->
+				<div id="col-right">
+					<div class="col-wrap">
+						<?php $this->list_table->display(); ?>
```
renatonascalves, line 374: I think one possible feature request we might get is to show the total number of users that read/dismissed the notice.

CleanShot 2024-09-02 at 13 16 09@2x

renatonascalves (changes requested):
✅  I'm done reviewing it! =D

src/bp-members/classes/class-bp-members-notices-rest-controller.php
```diff
@@ -0,0 +1,1090 @@
...
+
+		// The notice we just created will be active.
+		$notice        = bp_members_get_notice( $id );
+		$fields_update = $this->update_additional_fields_for_object( $notice, $request );
+
+		if ( is_wp_error( $fields_update ) ) {
+			return $fields_update;
+		}
+
+		$retval   = $this->prepare_item_for_response( $notice, $request );
```
renatonascalves, line 348: This method already sets `rest_ensure_response`. 🤔

renatonascalves: Do you know why we set it again in line 349?

src/bp-members/classes/class-bp-members-notices-rest-controller.php
```diff
@@ -0,0 +1,1090 @@
...
+	 * Update a notice.
+	 *
+	 * @since 15.0.0
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 * @return WP_REST_Response|WP_Error
+	 */
+	public function update_item( $request ) {
+		// Setting context.
+		$request->set_param( 'context', 'edit' );
```
renatonascalves, line 397: I think this should be passed by the consumer of the API. We have been removing this hardcoded values in other REST API endpoints.

src/bp-members/classes/class-bp-members-notices-rest-controller.php
```diff
@@ -0,0 +1,1090 @@
...
+	public function dismiss_notice_permissions_check( $request ) {
+		$retval = new WP_Error(
+			'bp_rest_authorization_required',
+			__( 'Sorry, you are not allowed to dismiss notices.', 'buddypress' ),
+			array(
+				'status' => rest_authorization_required_code(),
+			)
+		);
+
+		if ( is_user_logged_in() ) {
```
renatonascalves, line 545: I _think_ it is technically possible for _any_ logged in user to dismiss someone else's notices. As long as they have access to the notice ID. Is this expected?

src/bp-members/classes/class-bp-members-notices-rest-controller.php
```diff
@@ -0,0 +1,1090 @@
...
+	 * Delete a notice.
+	 *
+	 * @since 15.0.0
+	 *
+	 * @param WP_REST_Request $request Full details about the request.
+	 * @return WP_REST_Response|WP_Error
+	 */
+	public function delete_item( $request ) {
+		// Setting context.
+		$request->set_param( 'context', 'edit' );
```
renatonascalves, line 570: See my feedback about setting this manually.

src/bp-members/classes/class-bp-members-notices-rest-controller.php
```diff
@@ -0,0 +1,1090 @@
...
+
+	/**
+	 * Get the query params for notices collection.
+	 *
+	 * @since 15.0.0
+	 *
+	 * @return array
+	 */
+	public function get_collection_params() {
+		$params                       = parent::get_collection_params();
```
renatonascalves, line 941: This is adding some fields which I'm not sure are supported currently. Like `search`.

src/bp-members/classes/class-bp-members-notices-rest-controller.php
```diff
@@ -0,0 +1,1090 @@
...
+		);
+
+		$params['target'] = array(
+			'description'       => __( 'Limit result set to items concerning one or more specific targets.', 'buddypress' ),
+			'type'              => 'array',
+			'items'             => array(
+				'enum' => array( 'community', 'contributors', 'admins' ),
+				'type' => 'string',
+			),
+			'sanitize_callback' => 'wp_parse_list',
```
renatonascalves, line 987:
```suggestion
			'sanitize_callback' => 'bp_rest_sanitize_string_list',
```

src/bp-members/classes/class-bp-members-notices-rest-controller.php
```diff
@@ -0,0 +1,1090 @@
...
+			'validate_callback' => 'rest_validate_request_arg',
+		);
+
+		$params['priority'] = array(
+			'description'       => __( 'Limit result set to items having a specific priority.', 'buddypress' ),
+			'type'              => 'integer',
+			'items'             => array(
+				'enum' => array( 1, 2, 3 ),
```
renatonascalves, line 995: So it is not possible to fetch other types of items as an admin?

src/bp-members/classes/class-bp-members-notices-rest-controller.php
```diff
@@ -0,0 +1,1090 @@
...
+	 */
+	protected function prepare_item_for_database( $request ) {
+		$schema        = $this->get_item_schema();
+		$notice_id     = $request->get_param( 'id' );
+		$prepared_item = bp_members_get_notice( $notice_id );
+
+		// Notice ID.
+		if ( ! empty( $schema['properties']['id'] ) && ! empty( $prepared_item->id ) ) {
+			$prepared_item->id = $prepared_item->id;
+		}
```
renatonascalves, line 1028: Oddly, if the `id` property is empty, we are passing it along anyway. So I'm not sure how useful this conditional actually is.

src/bp-messages/classes/class-bp-rest-sitewide-notices-endpoint.php
```diff
@@ -0,0 +1,38 @@
...
+ */
+
+defined( 'ABSPATH' ) || exit;
+
+_deprecated_file(
+	basename( __FILE__ ),
+	'15.0.0',
+	'bp-members/classes/class-bp-members-notices-rest-controller.php',
+	esc_html__( 'The v1 of the BP REST API is deprecated, please use v2.', 'buddypress' )
+);
```
renatonascalves, line 17: I'm a but confused about having this class here. If the BP REST plugin is active, I'd expect we would get a PHP error since we would be loading the same class twice.

src/js/blocks/bp-members/webpack.config.js
```diff
@@ -10,12 +10,22 @@ module.exports = {
     ...defaultConfig,
 	...{
         entry: {
+			'notices-center/controller': {
```
renatonascalves, line 13: It seems there is a spacing issue here.

renatonascalves (changes requested):

renatonascalves:
> I remember having discussions with @renatonascalves about the best way to deprecate BP REST API v1 and class naming for the BP REST API v2. I was confronted to it and after more thoughts, I believe we should name v2 REST API Class controllers like this BP_{component}_{feature}_REST_Controller (I probably changed my mind there, sorry!): one of benefit for this is we are following the class autoloader regular map and don’t need to add entries for irregular ones. Here’s the deprecation strategy I used:

Changing the format of the files for the V2 endpoints is no problem at all. I think the main issue before was renaming the files from V1, but since we will bundle it in the bp rest plugin, keeping their names, again, it's no problem at all.

imath:
Hi @renatonascalves thanks a lot for your review 😍 . I'll look at it in details asap!

src/bp-members/classes/class-bp-members-notices-admin.php
imath: I agree, I thought about it myself 👍 . I'll see what I can do about it once most of the PR has been committed 😉

src/bp-members/classes/class-bp-members-notices-rest-controller.php
imath: I haven't changed how it currently is in https://github.com/buddypress/BP-REST/blob/master/includes/bp-messages/classes/class-bp-rest-sitewide-notices-endpoint.php#L305

Do you remember why you chose to force the context in the first place?

src/bp-members/classes/class-bp-members-notices-rest-controller.php
imath: Yes I have the same question: any idea why you set it this way in the first place?
See https://github.com/buddypress/BP-REST/blob/master/includes/bp-messages/classes/class-bp-rest-sitewide-notices-endpoint.php#L541

src/bp-members/classes/class-bp-members-notices-rest-controller.php
imath: Absolutely 😄

src/bp-members/classes/class-bp-members-notices-rest-controller.php
imath: I have no idea to be honest, I haven't touched this part from the BP REST plugin version 🤔

src/bp-members/classes/class-bp-members-notices-admin.php
imath: Good catch! I'll improve this.

src/bp-members/classes/class-bp-members-notices-rest-controller.php
imath: Notices were targeting every member of the community in the previous implementation, I agree I now need to be me restrictive here to check admin notices can only be viewed by admins and contributor notices can only be viewed by contributors.

src/bp-members/classes/class-bp-members-notices-rest-controller.php
imath: Good catch, I'll clean up these fields.

src/bp-members/bp-members-cache.php
```diff
@@ -125,3 +125,22 @@ function bp_members_reset_signup_cache_incrementor() {
...
+		$user_id = bp_loggedin_user_id();
+	}
+
+	wp_cache_delete( $user_id, 'bp_member_first_active_notice' );
+	wp_cache_delete( $user_id, 'bp_member_dismissed_notices' );
+	wp_cache_delete( $user_id, 'bp_member_notices_count' );
+	wp_cache_delete( $user_id, 'bp_member_top_priority_notices' );
+}
+add_action( 'bp_members_notice_after_save', 'bp_members_clear_user_notices_cache', 10, 0 );
+add_action( 'bp_members_notice_after_delete', 'bp_members_clear_user_notices_cache', 10, 0 );
```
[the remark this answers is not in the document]
imath: Actually no, because these to hooks are not returning the user ID as their first argument, reason why I used `0`.

src/bp-members/bp-members-notices.php
```diff
@@ -0,0 +1,1510 @@
...
+ */
+function bp_get_notice_content( $notice = null, $raw = false ) {
+	$notice_content = '';
+	$parsed_content = bp_get_parsed_notice_block( $notice );
+
+	if ( ! empty( $parsed_content['innerHTML'] ) ) {
+		$raw_content = $parsed_content['innerHTML'];
+
+		if ( $raw ) {
+			return $raw_content;
```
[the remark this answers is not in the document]
imath: Yes it's intentional 😄

src/bp-members/bp-members-notices.php
```diff
@@ -0,0 +1,1510 @@
...
+ * Output the selected target for the notice.
+ *
+ * @since 15.0.0
+ *
+ * @param BP_Members_Notice|null $notice The notice object.
+ */
+function bp_notice_target( $notice = null ) {
+	$target_key = bp_get_notice_target( $notice );
+
+	$targets = array(
```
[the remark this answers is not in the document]
imath: Good point, we can always improve it once most of the code is merged into Core 😅

src/bp-members/classes/class-bp-members-notice.php
imath: Most of these `::get` are performing 2 queries: one to list objects, the other to get total, I believe using it this way is avoiding making 2 queries when we only need one 😄

src/bp-members/classes/class-bp-members-notice.php
imath: I agree, let's do it once it's merged 👍

src/js/blocks/bp-members/webpack.config.js
imath: I've checked, it doesn't look it does 🤔

src/bp-core/classes/class-bp-admin.php
```diff
@@ -1687,28 +1653,111 @@ public function display_addons_table() {
...
+			if ( isset( $_GET['bp-dismissed'] ) ) {
+				$class    = empty( $_GET['bp-dismissed'] ) ? 'error' : 'updated';
+				$feedback = __( 'Notice successfully dismissed', 'buddypress' );
+
+				if ( 'error' === $class ) {
+					$feedback = __( 'The notice could not be dismissed', 'buddypress' );
+				}
+
+				printf(
+					'<div id="message" class="%1$s notice is-dismissible"><p>%2$s</p></div>',
```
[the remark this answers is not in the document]
imath: We are using the WordPress `id` to inherit admin styles, that's why we don't add a `bp-` prefix. I've checked in other Admin files.

src/bp-messages/classes/class-bp-rest-sitewide-notices-endpoint.php
imath: We had a discussion about it, the important part is to remove the deprecated class from the class autoloader irregular map.

src/bp-core/admin/bp-core-admin-schema.php
```diff
@@ -632,3 +626,61 @@ function bp_core_install_nonmember_opt_outs() {
...
 }
+
+/**
+ * Members Notices installation function.
+ *
+ * If the `bp_messages_notices` table exists, we simply need to rename it.
+ *
+ * @since 15.0.0
+ */
+function bp_core_install_members_notices() {
```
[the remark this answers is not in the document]
imath: It would be useful only if the `wp_bp_messages_notices` table still exists, the upgrade routine is renaming it to `wp_bp_notices`. I'm unsure it's worthing the effort, but let's get team's opinion about it!

src/bp-members/classes/class-bp-members-notices-admin.php
imath: user-feedback

When title and/or content are not provided it now displays a more meaningful error + highlights the required fields like WordPress does for categories/tags.

src/bp-members/classes/class-bp-members-notices-rest-controller.php
imath: I've updated this part this way:
```php
if ( is_user_logged_in() ) {
		// Users can only get notices targeting them.
		if ( 'view' === $context && bp_loggedin_user_id() === (int) $request->get_param( 'user_id' ) ) {
```

src/bp-members/classes/class-bp-members-notice.php
imath: I agree we should probably move to this syntax, but then we should do it for all BuddyPress objects. I think it's best to deal with it globally from another ticket.

src/bp-members/classes/class-bp-members-notice.php
imath: I prefer to keep the order/orderby readonly for now. Notices are always ordered by priority and date.

```php
ORDER BY priority ASC, date_sent DESC
```

src/bp-messages/classes/class-bp-rest-sitewide-notices-endpoint.php
imath: I've removed the deprecated class from the the autoloader.

src/bp-members/classes/class-bp-members-notices-rest-controller.php
imath: I've edited it to remove the ID if the schema doesn't have the ID property, I guess it's to prepare a create request.

src/bp-members/classes/class-bp-members-notices-rest-controller.php
imath: Only the `search` field needed to be removed ✅

imath:
@renatonascalves I believe I've improved most important parts according to your review.

About the 2 remaining REST API points: duplicate usage of `rest_ensure_response` and forcing the `edit` context for create/update/delete requests, I'll do what you confirm is best. I've copied these from the current v1 REST API endpoint.

About the other improvements + adding more unit tests, I'd suggest to work on it once the current code is merged 😅

src/bp-members/classes/class-bp-members-notices-rest-controller.php
renatonascalves: I **incorrectly** assumed that any `WP_REST_Server::EDITABLE` request use the `edit` context, which is not always the case. It's up to the API consumer to decide which data to show.

src/bp-members/classes/class-bp-members-notices-rest-controller.php
imath: I checked all other BP REST API endpoints as well as the `WP_REST_Posts_Controller`: using the `rest_ensure_response()` twice (inside the `prepare_item_for_response()` & into the request callback) is also practiced. I don't know why, but it looks to be the way to do it... So let's carry on using it twice! As this [function](https://developer.wordpress.org/reference/functions/rest_ensure_response/) is only making sure a `WP_REST_Response` object is returned, it's hurting noone: twice is always better than never!

imath:
Just updated the PR, it's now safe to test it again!

You don't need the following tip anymore as the v2 of the BP REST API has been merged into Trunk.

```php
function force_notices_feature_rest_api( $feature_class ) {
    add_action( 'bp_rest_api_init', array( $feature_class, 'rest_api_init' ), 10 );
}
add_action( 'bp_members_notices_setup_actions', 'force_notices_feature_rest_api' );
```
If you added it ^^ to test before, please remove it!

imath:
To avoid the theme issue @emaralive shared during latest dev-chats, I've just introduced a filter to keep the good old way to display notifications on the front end of the site.

Using:
```php
add_filter( 'bp_members_admin_bar_use_notifications_menu', '__return_true' );
```

Here's how it looks when you only have notifications
Capture d’écran 2024-10-12 à 11 55 11

Here's how it looks if you have a least one notice
Capture d’écran 2024-10-12 à 11 55 40

src/bp-members/bp-members-adminbar.php
```diff
@@ -165,106 +164,182 @@ function bp_members_admin_bar_user_admin_menu() {
...
+		);
+
+		if ( $notices_count ) {
+			$alert_class = array( 'pending-count', 'alert', 'admin-type' );
+			$notices     = array(
+				(object) array(
+					'id'      => 'notices',
+					'content' => sprintf(
+						/* translators: %s: total notices count */
+						_n( 'Warning: you have %s unread notice', 'Warning: you have %s unread notices', $notices_count, 'buddypress' ), number_format_i18n( $notices_count )
```
dcavins, line 205: "Warning" is a little too strong of a word. How about "Important: You have %s unread notice"? (The American English style is to often capitalize the first word after a colon if the phrase is a sentence.)

src/bp-core/deprecated/15.0.php
```diff
@@ -0,0 +1,391 @@
...
+
+/**
+ * Invalidate cache for notices.
+ *
+ * @since 2.0.0
+ * @deprecated 15.0.0
+ */
+function bp_notices_clear_cache() {
+	_deprecated_function( __FUNCTION__, '15.0.0' );
+}
```
dcavins, line 391: Wow! So many deprecations. Are we reasonably confident that themes haven't built their own notices interface using these functions?

src/bp-members/bp-members-notices.php
```diff
@@ -0,0 +1,1515 @@
...
+
+	$results = array();
+	foreach ( $keys as $key ) {
+		$results[] = delete_metadata( 'notice', $notice_id, $key, $meta_value, $delete_all );
+	}
+	$result = array_filter( $results );
+
+	remove_filter( 'query', 'bp_filter_metaid_column_name' );
+
+
```
dcavins, line 175: Extra line return here.

src/bp-members/classes/class-bp-members-notice.php
```diff
@@ -0,0 +1,760 @@
...
+		if ( false === $dismissed ) {
+			$dismissed = self::get(
+				array(
+					'user_id'   => $user_id,
+					'dismissed' => true,
+					'fields'    => 'ids',
+				)
+			);
+
+			// Cache user's first active notice.
```
dcavins, line 697: Cache user's dismissed notices.

dcavins:
This is an excellent change, and one that I appreciated more after seeing how it improves the focus of the existing messages and notifications code. Wow, what a lot of work! Thanks for doing this! It should make all of these somewhat related pieces easier to maintain and improve. :)

emaralive:
> To avoid the theme issue @emaralive shared during latest dev-chats, I've just introduced a filter to keep the good old way to display notifications on the front end of the site.
> 
> Using:
> 
> ```
> add_filter( 'bp_members_admin_bar_use_notifications_menu', '__return_true' );
> ```
> 
> Here's how it looks when you only have notifications Capture d’écran 2024-10-12 à 11 55 11
> 
> Here's how it looks if you have a least one notice Capture d’écran 2024-10-12 à 11 55 40

The filter attempts to apply a _band-aid_ ( :arrow_left:  for lack of a better term) as a solution for the issue that can be seen in the following image (same image shown in slack when I raised the issue:

Git-WP-Built-by-Photographers-for-Photographers-10-02-2024_12_52_PM

However, the indicated filter assumes the toolbar is available on the front-end which is not the case for all sites, as can be demonstrably seen. Applying the filter, in the case of "**no toolbar**" looks like the following (no indication or access to notifications because the toolbar is not present):

bp_members_admin_bar_use_notifications_menu-true

If a user has access to their profile area within wp-admin, they can turn off the toolbar or the site  admin can turn off the toolbar universally with:

```
add_filter( 'show_admin_bar', '__return_false' );
 ```

What's also interesting is that when the "**show_admin_bar**" filter returns a **false** or when the "**Show Toolbar when viewing site**" checkbox is disabled within a user's profile (wp-admin) there are no **notices** on the front-end, regardless of whether the "**bp_members_admin_bar_use_notifications_menu**" filter returns a **true**. This infers that a dependency exists of some sorts and TBH, I've not examined the code, this is just testing various scenarios with the **TwentyEleven** theme.  Another bit of information, the reason notices can be seen when there appears to be "**no toolbar**" is due to CSS is being used to hide the **toolbar**.

A final note for this comment is that there is an additional issue when the "**bp_is_members_notices_active**" returns a **false**, as can be seen in this next image:

bp_is_members_notices_active-false

More testing is required with other themes and IMHO, this isn't ready to be released with 15.0.

PR #371, opened 2024-09-04
Imprved the checking for BP Class Plugin

narenin:
Trac ticket: https://buddypress.trac.wordpress.org/ticket/9181

---

### DIFF
--- a/src/bp-core/bp-core-functions.php
+++ b/src/bp-core/bp-core-functions.php
@@ -5295,8 +5295,5 @@ function bp_is_admin( $screen_id = '' ) {
  * @return bool True if bp_classic is active, false otherwise.
  */
 function is_bp_classic_active() {
-    if ( function_exists( 'bp_classic' ) || is_plugin_active( 'bp-classic/bp-classic.php' ) ) {
-        return true;
-    }
-    return false;
+	return function_exists( 'bp_classic' ) || is_plugin_active( 'bp-classic/bp-classic.php' );
 }
--- a/src/bp-core/bp-core-functions.php
+++ b/src/bp-core/bp-core-functions.php
@@ -5295,7 +5295,7 @@ function bp_is_admin( $screen_id = '' ) {
  * @return bool True if bp_classic is active, false otherwise.
  */
 function is_bp_classic_active() {
-    if ( function_exists( 'bp_classic' ) || ( function_exists( 'is_plugin_active' ) && is_plugin_active( 'bp-classic/bp-classic.php' ) ) ) {
+    if ( function_exists( 'bp_classic' ) || is_plugin_active( 'bp-classic/bp-classic.php' ) ) {
         return true;
     }
     return false;
--- a/src/bp-core/bp-core-functions.php
+++ b/src/bp-core/bp-core-functions.php
@@ -5295,5 +5295,5 @@ function bp_is_admin( $screen_id = '' ) {
  * @return bool True if bp_classic is active, false otherwise.
  */
 function bp_is_classic() {
-	return function_exists( 'bp_classic' ) || is_plugin_active( 'bp-classic/bp-classic.php' );
+	return function_exists( 'bp_classic' ) || ( function_exists( 'is_plugin_active' ) && is_plugin_active( 'bp-classic/bp-classic.php' ) );
 }
--- a/src/bp-core/bp-core-functions.php
+++ b/src/bp-core/bp-core-functions.php
@@ -5286,3 +5286,17 @@ function bp_is_admin( $screen_id = '' ) {
 
 	return $screen_id === $bp->admin->current_screen;
 }
+
+/**
+ * Common function to check if bp_classic is active or not.
+ *
+ * @since 15.0.0
+ *
+ * @return bool True if bp_classic is active, false otherwise.
+ */
+function is_bp_classic_active() {
+    if ( function_exists( 'bp_classic' ) || ( function_exists( 'is_plugin_active' ) && is_plugin_active( 'bp-classic/bp-classic.php' ) ) ) {
+        return true;
+    }
+    return false;
+}
--- a/src/bp-core/bp-core-functions.php
+++ b/src/bp-core/bp-core-functions.php
@@ -5294,6 +5294,6 @@ function bp_is_admin( $screen_id = '' ) {
  *
  * @return bool True if bp_classic is active, false otherwise.
  */
-function is_bp_classic_active() {
+function bp_is_classic() {
 	return function_exists( 'bp_classic' ) || is_plugin_active( 'bp-classic/bp-classic.php' );
 }
--- a/src/bp-templates/bp-nouveau/buddypress-functions.php
+++ b/src/bp-templates/bp-nouveau/buddypress-functions.php
@@ -122,7 +122,7 @@ class BP_Nouveau extends BP_Theme_Compat {
 			);
 
 			// When BP Classic is activated, regular themes need this filter. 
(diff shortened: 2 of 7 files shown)


src/bp-templates/bp-nouveau/buddypress-functions.php
```diff
@@ -122,7 +122,7 @@ function() {
 			);
 
 			// When BP Classic is activated, regular themes need this filter. 
-			if ( function_exists( 'bp_classic' ) ) {
+			if ( function_exists( 'bp_classic' ) || is_plugin_active( 'bp-classic/bp-classic.php' ) || is_plugin_active( 'buddyclassic/bp-classic.php' ) ) {
```
renatonascalves, line 125: Is `buddyclassic` a real plugin?

narenin: @renatonascalves Thanks for reviewing this PR. I have added this just for the case when user will rename folder to `buddyclassic` from `bp-classic`

narenin: @renatonascalves Please let me know if I need to made any other changes in this.

renatonascalves: I'd remove the buddyclassic part, actually. It is not a _real_ plugin.

renatonascalves, line 125: There are many places where we are doing similar checks that we should update too: https://github.com/search?q=org%3Abuddypress+function_exists(+%27bp_classic%27+)&type=Code

Maybe create a new function to check if the BP Classic plugin is active?! And reuse it where applicable.

renatonascalves (changes requested):

src/bp-templates/bp-nouveau/buddypress-functions.php
narenin: @renatonascalves Thanks for the detailed feedback, I have implemented the suggestions can you please take a look.

narenin: @renatonascalves @imath Please let me know if there is any feedback for me.

src/bp-core/bp-core-functions.php
```diff
@@ -5286,3 +5286,17 @@ function bp_is_admin( $screen_id = '' ) {
...
+ *
+ * @since 15.0.0
+ *
+ * @return bool True if bp_classic is active, false otherwise.
+ */
+function is_bp_classic_active() {
+    if ( function_exists( 'bp_classic' ) || is_plugin_active( 'bp-classic/bp-classic.php' ) ) {
+        return true;
+    }
+    return false;
```
imath, line 5301:
```suggestion
    return function_exists( 'bp_classic' ) || is_plugin_active( 'bp-classic/bp-classic.php' );
```
I believe we can save 2 lines of code 😅

imath (commented):
Hi @narenin thanks for your contribution.

I wonder what's fastest, if checking if BP Classic is active using the WordPress function and since this seems required we could probably remove the `function_exists` check?

src/bp-core/bp-core-functions.php
narenin: @imath Thanks for reviewing this. I have implemented the suggestions please check.

src/bp-core/bp-core-functions.php
```diff
@@ -5286,3 +5286,14 @@ function bp_is_admin( $screen_id = '' ) {
...
 }
+
+/**
+ * Common function to check if bp_classic is active or not.
+ *
+ * @since 15.0.0
+ *
+ * @return bool True if bp_classic is active, false otherwise.
+ */
+function is_bp_classic_active() {
```
renatonascalves, line 5297: One last suggestion to start with our custom prefix: `bp_`
```suggestion
function bp_is_classic() {
```

renatonascalves (changes requested):

src/bp-core/bp-core-functions.php
narendraie: @renatonascalves Thanks for the suggestion, I have implemented the same please check.

narenin: @renatonascalves Please let me know if any other changes needs to be made.

renatonascalves (approved):
All good! 🎈

src/bp-core/bp-core-functions.php
```diff
@@ -5286,3 +5286,14 @@ function bp_is_admin( $screen_id = '' ) {
...
+
+/**
+ * Common function to check if bp_classic is active or not.
+ *
+ * @since 15.0.0
+ *
+ * @return bool True if bp_classic is active, false otherwise.
+ */
+function bp_is_classic() {
+	return function_exists( 'bp_classic' ) || is_plugin_active( 'bp-classic/bp-classic.php' );
```
imath, line 5298: All PHP Unit tests are failing, we need more work about this PR. For a reason I ignore for now the `is_plugin_active()` function is not yet available when doing these tests.

`PHP Fatal error:  Uncaught Error: Call to undefined function is_plugin_active() in /tmp/wordpress/wp-content/plugins/buddypress/src/bp-core/bp-core-functions.php:5298`

imath (changes requested):
PHP Unit tests are not good 😓 , let's understand why!

src/bp-core/bp-core-functions.php
narenin: @imath Thanks for the review and feedback, should we add something like this to `tests/phpunit/bootstrap.php` to handle that fatal error?

`if ( ! function_exists( 'is_plugin_active' ) ) {
    require_once ABSPATH . 'wp-admin/includes/plugin.php';
}
`

renatonascalves: I _think_ `is_plugin_active` is not loaded for front-end requests, only `wp-admin` requests. So when we load `bp-nouveau`, we make use of this function and it is complaining it doesn't exist.

```php
if ( ! function_exists( 'is_plugin_active' ) ) {
  require_once ABSPATH . 'wp-admin/includes/plugin.php';
} 
```

^ this seems to be the solution suggested [here](https://developer.wordpress.org/reference/functions/is_plugin_active/#comment-616) for front-end support too.

src/bp-core/bp-core-functions.php
```diff
@@ -5286,3 +5286,14 @@ function bp_is_admin( $screen_id = '' ) {
...
+
+/**
+ * Common function to check if bp_classic is active or not.
+ *
+ * @since 15.0.0
+ *
+ * @return bool True if bp_classic is active, false otherwise.
+ */
+function bp_is_classic() {
+	return function_exists( 'bp_classic' ) || ( function_exists( 'is_plugin_active' ) && is_plugin_active( 'bp-classic/bp-classic.php' ) );
```
renatonascalves, line 5298: I'm under the impression that @imath suggested to add `is_plugin_active` to this file. Like a polyfill. 🤔 

https://github.com/buddypress/buddypress/blob/master/src/bp-core/bp-core-wpabstraction.php#L190%7CL203

Although, your approach works too. So I guess the question is if we wanna polyfill `is_plugin_active `

renatonascalves (changes requested):

src/bp-core/bp-core-functions.php
narenin: @imath @renatonascalves Thanks for the feedback, I have implemented the feedback, could you please take the look.

PR #384, opened 2024-10-11
Improve deleted users messages in the web and REST API

renatonascalves:
Trac ticket: https://buddypress.trac.wordpress.org/ticket/9160

---


src/bp-templates/bp-nouveau/includes/messages/ajax.php
```diff
@@ -420,7 +420,7 @@ function bp_nouveau_ajax_get_user_message_threads() {
 	// Remove the bp_current_action() override.
```
imath, line 420: According to ticket's comment (`#30`) I believe you might need to add something like this:

[9160.patch](https://github.com/user-attachments/files/17347335/9160.patch)

imath (changes requested):
I've added a patch for BP Nouveau's messages Ajax, you may need it to be sure the Backbone UI outputs consistent infos about deleted users.

imath (approved):
Looks good. I've tested it: works great.

PR #389, opened 2024-10-20
Patch/9207

imath:
To fix WP Playground issues, I suggest to:

- Send the current canonical URL into Ajax request vars. `wp_get_referer()` is not set which explains why BP URI globals are not set into Ajax context by WP Playground.
- Ignore `$_COOKIE['bp-message']` if its value is set to `deleted`. It looks like the cookie is not deleted but its value is set to `deleted` once it's removed by BuddyPress.
- About the ` SIGNED` conversion or casting SQL issue, it looks like it's a SQLlite issue that will be hard to cope with: see https://wordpress.org/support/topic/wordpress-meta-query-not-working/ I used a workaround to get rid of it for now.

> [!IMPORTANT]
> I'll need to run `grunt commit` before SVN committing this PR.

Trac ticket: https://buddypress.trac.wordpress.org/ticket/9207

---

### DIFF
--- a/src/bp-xprofile/classes/class-bp-xprofile-group.php
+++ b/src/bp-xprofile/classes/class-bp-xprofile-group.php
@@ -352,10 +352,12 @@ class BP_XProfile_Group {
 			}
 		}
 
+		$signup_fields_order = bp_xprofile_get_signup_field_ids();
+
 		// Pull field objects from the cache.
 		$fields = array();
 		foreach ( $field_ids as $field_id ) {
-			if ( true === $r['signup_fields_only'] && ! in_array( $field_id, bp_xprofile_get_signup_field_ids(), true ) ) {
+			if ( true === $r['signup_fields_only'] && ! in_array( $field_id, $signup_fields_order, true ) ) {
 				continue;
 			}
 
--- a/src/bp-xprofile/classes/class-bp-xprofile-group.php
+++ b/src/bp-xprofile/classes/class-bp-xprofile-group.php
@@ -352,12 +352,10 @@ class BP_XProfile_Group {
 			}
 		}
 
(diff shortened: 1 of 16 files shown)


src/bp-xprofile/classes/class-bp-xprofile-group.php
```diff
@@ -352,12 +352,10 @@ public static function get( $args = array() ) {
 			}
 		}
 
-		$signup_fields_order = bp_xprofile_get_signup_field_ids();
-
 		// Pull field objects from the cache.
 		$fields = array();
 		foreach ( $field_ids as $field_id ) {
-			if ( true === $r['signup_fields_only'] && ! in_array( $field_id, $signup_fields_order, true ) ) {
+			if ( true === $r['signup_fields_only'] && ! in_array( $field_id, bp_xprofile_get_signup_field_ids(), true ) ) {
```
renatonascalves, line 358: This was changed here to avoid calling the object cache, or the database, repeatly in the foreach loop.

https://github.com/buddypress/buddypress/commit/15f403f578fed2ee35f3967f50e5510d57395c0a

renatonascalves (changes requested):

src/bp-xprofile/classes/class-bp-xprofile-group.php
renatonascalves: Is this breaking something?

imath: Not really, thanks for your comment. As It's generating an error with WP Playground's SQLite db, I was a bit surprised this request was performed even if not on the registration page. I'll try to improve this part differently 👍

PR #403, opened 2024-11-01
Start using `BP_Component_Feature` to organize the Members Invitations feature

imath:
Stop using `BP_Component` in favor of `BP_Component_Feature` to avoid messing with BuddyPress component globals and "faking" a component to register the feature navigation.

Trac ticket: https://buddypress.trac.wordpress.org/ticket/9254

---

### DIFF
--- a/src/bp-core/classes/class-bp-component-feature.php
+++ b/src/bp-core/classes/class-bp-component-feature.php
@@ -1,0 +1,223 @@
+<?php
+/**
+ * BuddyPress Component's feature Class.
+ *
+ * @package BuddyPress
+ * @subpackage Core
+ * @since 15.0.0
+ */
+
+// Exit if accessed directly.
+defined( 'ABSPATH' ) || exit;
+
+/**
+ * BuddyPress Component Feature API.
+ *
+ * @since 15.0.0
+ */
+class BP_Component_Feature extends BP_Component {
+
+	/**
+	 * Unique ID for the feature.
+	 *
+	 * @since 15.0.0
+	 *
+	 * @var string
+	 */
+	public $id = '';
+
+	/**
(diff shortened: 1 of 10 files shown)


src/bp-members/classes/class-bp-members-invitations-feature.php
```diff
@@ -0,0 +1,188 @@
...
+		parent::init( 'invitations', 'members' );
+	}
+
+	/**
+	 * Set up Signups feature global variables.
+	 *
+	 * @since 15.0.0
+	 */
+	public function globals() {
+		buddypress()->members->invitations = new stdClass();
```
dcavins, line 38: This class seems to be only used to store the `query_loop` as used in `bp-members-template.php`. Would it make more sense to fold it into the `members->active_features->invitations` object in the future?

src/bp-core/classes/class-bp-component-feature.php
```diff
@@ -0,0 +1,223 @@
...
+	 * This is a component's feature.
+	 *
+	 * @since 15.0.0
+	 *
+	 * @var boolean
+	 */
+	public $is_feature = true;
+
+	/**
+	 * Disable the `BP_Component()` starting method.
```
dcavins, line 57: I think I understand that you're doing this to help differentiate features from components and essentially have two distinct setup routines. Maybe the note here should say explicitly that classes that extend BP_Component_Feature should use the init() method and classes that extend BP_Component should use start(). I'm not sold on this wording, but this section of the code confused me at first look, so I wonder if we can help code readers.

dcavins (approved):
Overall, this is a great improvement for making sense of how these pieces fit together. Having an API-style way to register features of components will really help us in the long run. Thanks!

PR #411, opened 2025-06-04
Fix: bp_get_blog_latest_post using the_title filters with missing param

rollybueno:
Reported on https://buddypress.trac.wordpress.org/ticket/9286, the bp_get_blog_latest_post() will be using the_title filters but only provided 1 out of 2 **required** params

The patch includes creating new function called bp_get_blog_latest_post_id() which is just a copy of bp_get_blog_latest_post_title(), except it will return latest post ID instead of title.

Then we use this for the  required param on the_title filters as per documentation: https://developer.wordpress.org/reference/hooks/the_title/

Ticket: https://buddypress.trac.wordpress.org/ticket/9286

Filter docs: https://developer.wordpress.org/reference/hooks/the_title/

Trac ticket: https://buddypress.trac.wordpress.org/ticket/9286

---

### DIFF
--- a/src/bp-blogs/bp-blogs-template.php
+++ b/src/bp-blogs/bp-blogs-template.php
@@ -753,7 +753,8 @@ function bp_blog_latest_post( $args = array() ) {
 			)
 		);
 
-		$retval = bp_get_blog_latest_post_title();
+		$retval  = bp_get_blog_latest_post_title();
+		$post_id = bp_get_blog_latest_post_id();
 
 		if ( ! empty( $retval ) ) {
 			if ( ! empty( $r['latest_format'] ) ) {
@@ -768,12 +769,12 @@ function bp_blog_latest_post( $args = array() ) {
 				$retval = sprintf(
 					/* translators: %s: the title of the latest post */
 					__( 'Latest Post: %s', 'buddypress' ),
-					'<a href="' . $blogs_template->blog->latest_post->guid . '">' . apply_filters( 'the_title', $retval ) . '</a>'
+					'<a href="' . $blogs_template->blog->latest_post->guid . '">' . apply_filters( 'the_title', $retval, $post_id ) . '</a>'
 				);
 			} else {
 
 				/** This filter is documented in bp-blogs/bp-blogs-template.php */
-				$retval = '<a href="' . $blogs_template->blog->latest_post->guid . '">' . apply_filters( 'the_title', $retval ) . '</a>';
+				$retval = '<a href="' . $blogs_template->blog->latest_post->guid . '">' . apply_filters( 'the_title', $retval, $post_id ) . '</a>';
 			}
 		}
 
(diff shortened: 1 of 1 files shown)


src/bp-blogs/bp-blogs-template.php
```diff
@@ -789,6 +790,45 @@ function bp_get_blog_latest_post( $args = array() ) {
...
 
+	/**
+	 * Output the ID of the latest post on the current blog in the loop.
+	 *
+	 * @since 14.3.4
+	 *
+	 * @see bp_get_blog_latest_post_id()
+	 */
+	function bp_blog_latest_post_id() {
+		echo esc_html( bp_get_blog_latest_post_id() );
```
dcavins, line 801: You shouldn't need `esc_html()` here, as long as `bp_get_blog_latest_post_id()` always returns an integer.

src/bp-blogs/bp-blogs-template.php
```diff
@@ -789,6 +790,45 @@ function bp_get_blog_latest_post( $args = array() ) {
...
+	}
+
+	/**
+	 * Return the ID of the latest post on the current blog in the loop.
+	 *
+	 * @since 14.3.4
+	 *
+	 * @global BP_Blogs_Template $blogs_template The main blog template loop class.
+	 *
+	 * @return string Post ID.
```
dcavins, line 811: The return type should be an integer, since a post ID is an integer.

src/bp-blogs/bp-blogs-template.php
```diff
@@ -789,6 +790,45 @@ function bp_get_blog_latest_post( $args = array() ) {
...
+		global $blogs_template;
+
+		$retval = '';
+
+		if ( ! empty( $blogs_template->blog->latest_post ) && ! empty( $blogs_template->blog->latest_post->ID ) ) {
+			$retval = $blogs_template->blog->latest_post->ID;
+		}
+
+		/**
+		 * Filters the title text of the latest post on the current blog in the loop.
```
dcavins, line 823: This filter doc block still refers to the title function filter.

src/bp-blogs/bp-blogs-template.php
```diff
@@ -789,6 +790,45 @@ function bp_get_blog_latest_post( $args = array() ) {
...
+		 *
+		 * @since 1.7.0
+		 *
+		 * @param string $retval Title text for the latest post.
+		 */
+		return apply_filters( 'bp_get_blog_latest_post_id', $retval );
```
dcavins, line 829: It shouldn't be needed, but maybe adding a cast to integer would make everyone feel better here. Like:
`return apply_filters( 'bp_get_blog_latest_post_id', (int) $retval );`

src/bp-blogs/bp-blogs-template.php
```diff
@@ -789,6 +790,45 @@ function bp_get_blog_latest_post( $args = array() ) {
...
+	 * @since 14.3.4
+	 *
+	 * @global BP_Blogs_Template $blogs_template The main blog template loop class.
+	 *
+	 * @return string Post ID.
+	 */
+	function bp_get_blog_latest_post_id() {
+		global $blogs_template;
+
+		$retval = '';
```
dcavins, line 816: The default should probably be `0` since we're returning an integer.

dcavins (commented):
Thanks for your PR!

src/bp-blogs/bp-blogs-template.php
rollybueno: Thanks for your feedback. I was get used to the Plugin Check, maybe it's became a habit now :sweat_smile: 

Corrected with the rest of your feedback. :beer:

PR #412, opened 2025-06-04
Bugfix/xprofile profiledata

rollybueno:
The exists() method was attempting to convert a WP_Error object to a string when
checking for profile data existence. This fix:

1. Initializes $retval as false by default
2. Properly handles WP_Error and empty results from database queries
3. Ensures consistent boolean return value throughout the method
4. Removes unnecessary boolean casting in apply_filters_ref_array

This prevents the "Object of class WP_Error could not be converted to string" error
while maintaining the expected boolean return value of the exists() method.

Trac ticket: https://buddypress.trac.wordpress.org/ticket/9290

---

### DIFF
--- a/src/bp-xprofile/classes/class-bp-xprofile-profiledata.php
+++ b/src/bp-xprofile/classes/class-bp-xprofile-profiledata.php
@@ -126,6 +126,7 @@ class BP_XProfile_ProfileData {
 	public function exists() {
 		global $wpdb;
 
+		// Check cache first.
 		$cache_key = "{$this->user_id}:{$this->field_id}";
 		$cached    = wp_cache_get( $cache_key, 'bp_xprofile_data' );
 		$retval    = false;
--- a/src/bp-xprofile/classes/class-bp-xprofile-profiledata.php
+++ b/src/bp-xprofile/classes/class-bp-xprofile-profiledata.php
@@ -126,15 +126,19 @@ class BP_XProfile_ProfileData {
 	public function exists() {
 		global $wpdb;
 
-		// Check cache first.
 		$cache_key = "{$this->user_id}:{$this->field_id}";
 		$cached    = wp_cache_get( $cache_key, 'bp_xprofile_data' );
+		$retval    = false;
 
 		if ( $cached && ! empty( $cached->id ) ) {
 			$retval = true;
 		} else {
 			$bp     = buddypress();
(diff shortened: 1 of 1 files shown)


src/bp-xprofile/classes/class-bp-xprofile-profiledata.php
```diff
@@ -126,15 +126,19 @@ public function populate( $field_id, $user_id ) {
...
-		// Check cache first.
 		$cache_key = "{$this->user_id}:{$this->field_id}";
 		$cached    = wp_cache_get( $cache_key, 'bp_xprofile_data' );
+		$retval    = false;
 
 		if ( $cached && ! empty( $cached->id ) ) {
 			$retval = true;
 		} else {
 			$bp     = buddypress();
 			$retval = $wpdb->get_row( $wpdb->prepare( "SELECT id FROM {$bp->profile->table_name_data} WHERE user_id = %d AND field_id = %d", $this->user_id, $this->field_id ) );
```
dcavins, line 138: I think you want to use another variable name here, like 
`$dbval = $wpdb->get_row( $wpdb->prepare( "SELECT id FROM {$bp->profile->table_name_data} WHERE user_id = %d AND field_id = %d", $this->user_id, $this->field_id ) );`
else you could be assigning a `WP_Error` to `$retval`, and the below check won't act on it (so WP_Error would still be passed to the filter).

src/bp-xprofile/classes/class-bp-xprofile-profiledata.php
```diff
@@ -126,15 +126,19 @@ public function populate( $field_id, $user_id ) {
...
+
+			if ( ! is_wp_error( $retval ) && ! empty( $retval ) ) {
```
dcavins, line 140: Also here, I think you want 
`if ( ! is_wp_error( $dbval ) && ! empty( $dbval ) ) {`
so you're only setting `$retval` to true (at line 140) when the database returns a valid value.

dcavins (commented):
Thanks for your pull request! I have a couple of comments that I've added to specific lines of code.

src/bp-xprofile/classes/class-bp-xprofile-profiledata.php
rollybueno: Thank you for your feedback. It makes sense to keep the $retval integrity intact on boolean. Updated the PR

PR #437, opened 2026-01-24
Activity: Add missing PHPDoc return descriptions for mention helpers

manhphuc:
This PR adds missing PHPDoc `@return` descriptions for two mention-related helper functions in:

- src/bp-activity/bp-activity-functions.php

The updated docblocks now accurately describe the boolean return values based on the actual behavior of the functions.

No functional changes.


renatonascalves:
This is unnecessary for a boolean.

